PR #25, opened 2026-05-14
⚡ Bolt: [performance improvement] Batched SQL query for sqlserver schema extraction

rfv-code:
💡 What: Modifies the schema extraction process for SQL Server to batch table columns in a single `IN (...)` query (chunked to a max of 1000 to respect SQL Server's parameter limit) rather than looping over `tables` and running an `await colReq.query()` for every single table.

🎯 Why: To fix an N+1 query problem during database schema extraction. Previously, if an application had 200 tables, this generated 201 sequential database queries to load metadata, making schema discovery quite slow.

📊 Impact: Reduces roundtrips significantly. The number of metadata extraction queries goes from $O(N)$ (where N is the number of tables) to $O(N/1000)$.

🔬 Measurement: `npm run test -- tests/connectors/db/drivers/test_sqlserver.test.ts` to ensure backward compatibility and see how execution avoids sequential per-table calls.

---
*PR created automatically by Jules for task [16147738040412730267](https://jules.google.com/task/16147738040412730267) started by @rfv-code*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -1,0 +1,3 @@
+## 2024-05-14 - Batch Schema Extraction
+**Learning:** For database schema extraction, querying columns table-by-table in a loop can cause an N+1 query problem, making it slow. The MySQL driver was optimized to use a batched query (`IN (?, ?, ...)`), but `sqlserver.ts` is still using a loop.
+**Action:** Optimize schema extraction in `sqlserver.ts` by using a single batched query to get columns for all tables, replacing the `for` loop over `tables` that executes a query for each.
--- a/src/connectors/db/lib/drivers/sqlserver.ts
+++ b/src/connectors/db/lib/drivers/sqlserver.ts
@@ -287,17 +287,33 @@ export class SqlServerDriver extends DatabaseDriver {
       const tableRes = await req.query(tableSql);
       const tables = tableRes.recordset ?? [];
 
-      const out: Table[] = [];
-      for (const row of tables) {
-        const tableName = String(
+      const tableNames: string[] = tables.map((row) =>
+        String(
           (row as Record<string, unknown>)["table_name"] ??
             (row as Record<string, unknown>)["TABLE_NAME"],
-        );
+        ),
+      );
+
+      if (tableNames.length === 0) return [];
+
+      // G-SCHEMA-BATCH: Avoid N+1 query problem by fetching column
+      // metadata for all tables in chunks of up to 1000 tables.
+      // The chunks are needed because SQL Server limits the number of
+      // parameters per query to 2100.
+      const colsByTable = new Map<string, Column[]>();
+      const chunkSize = 1000;
+
+      for (let chunkStart = 0; chunkStart < tableNames.length; chunkStart += chunkSize) {
+        const chunk = tableNames.slice(chunkStart, chunkStart + chunkSize);
+
         const colReq = handle.pool.request();
         colReq.input("schema", ns);
-        colReq.input("table", tableName);
+
+        const placeholders = chunk.map((_, i) => `@table${i}`).join(", ");
+        chunk.forEach((name, i) => colReq.input(`table${i}`, name));
+
         const colRes = await colReq.query(
-          "SELECT c.COLUMN_NAME AS column_name, c.DATA_TYPE AS data_type, " +
+          "SELECT c.TABLE_NAME AS table_name, c.COLUMN_NAME AS column_name, c.DATA_TYPE AS data_type, " +
             "c.IS_NULLABLE AS is_nullable, c.COLUMN_DEFAULT AS column_default, " +
             "CASE WHEN tc.CONSTRAINT_TYPE = 'PRIMARY KEY' THEN 1 ELSE 0 END AS is_pk " +
             "FROM INFORMATION_SCHEMA.COLUMNS c " +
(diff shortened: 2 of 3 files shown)


rfv-code (commented):
## Code review

**Verdict: Conditional approve.** The optimization pattern is correct and aligns with the existing MySQL implementation (`src/connectors/db/lib/drivers/mysql.ts:256`), but **test coverage for the changed paths is insufficient** and the PR description does not mention the mandatory invariant-path checks from CONTRIBUTING.md. I'd ask for additional tests and an eval-14/eval-20 manual run note before merge.

### Correctness checks (passed)

- **Mirrors MySQL's `G-SCHEMA-BATCH` pattern.** Same `colsByTable` grouping, same final emit loop preserving `tableNames` order, same per-table empty-column default (`colsByTable.get(tableName) ?? []`). Good.
- **Parameter limit math.** `1 (schema) + 1000 (chunk) = 1001` per query — comfortably under SQL Server's 2100. Chunk loop correctness: `slice(chunkStart, chunkStart + chunkSize)` is half-open and bounded by length; final chunk is correctly partial.
- **JOIN semantics preserved.** The LEFT JOINs to `KEY_COLUMN_USAGE` and `TABLE_CONSTRAINTS` still match per-row on `(TABLE_SCHEMA, TABLE_NAME, COLUMN_NAME)` and constraint name, so the per-column `is_pk` derivation is unchanged. Switching `WHERE TABLE_NAME = @table` to `WHERE TABLE_NAME IN (...)` keeps the schema-scoped filter via `c.TABLE_SCHEMA = @schema`.
- **`ORDER BY c.TABLE_NAME, c.ORDINAL_POSITION` added.** Correct — needed so columns stay in declaration order within each table after merging.
- **Empty short-circuit.** `if (tableNames.length === 0) return [];` is a behavior change from the prior version (which would have returned `[]` after an empty loop anyway), so functionally identical. Fine.

### Issues to address before merge

1. **Invariant-path PR hygiene unmet.** `src/connectors/db/lib/drivers/` is one of the four invariant paths called out in [`CONTRIBUTING.md`](CONTRIBUTING.md) (§ PR hygiene) and [`docs/architecture-invariants.md`](docs/architecture-invariants.md). Per the contributor docs:
   > If you touch one of the four invariant paths above, run doc-wiki's `eval-14` (`wiki-gather-credential-boundary`) and `eval-20` (`wiki-gather-db-postgres-success`) manually before merging.

   The PR description doesn't mention running these. Please run them and add the result (pass/fail + a sentence on what you observed) to the PR body or a comment. The eval suite has caught bugs in this exact area before — see `docs/architecture-invariants.md` §4 on the schema dispatch invariant.

2. **Test coverage gaps.** The only test update is adding `table_name: "products"` to two existing column rows. This is the bare minimum to keep the test passing. It does not cover the new code paths:

   - **Multi-table column bucketing.** Add a test with at least two distinct tables (e.g. `products` and `orders`), each contributing multiple column rows, asserting the columns end up grouped under the correct `Table` with correct ordering.
   - **`tableNames.length === 0` short-circuit.** Trivial test, but worth locking in — mock the `INFORMATION_SCHEMA.TABLES` query to return an empty recordset and assert `getSchemaAsync` returns `[]` without ever issuing the columns query.
   - **Chunking boundary.** Doesn't need a literal 1000-table test (slow), but a smaller unit-level test with `chunkSize` injection or a 2-table-per-chunk fixture would prove the chunk loop runs the right number of times and concatenates results. Even simpler: assert that `colReq.input("table0", ...)` and (for a 2-table case) `colReq.input("table1", ...)` were both called, and that the generated SQL contains `@table0, @table1`.
   - **Tables with no columns.** A table reported by `INFORMATION_SCHEMA.TABLES` but absent from the columns recordset should yield a `Table` with `columns: []` (the `colsByTable.get(...) ?? []` fallback). Lock this in.

3. **Test mock fragility.** The existing `getSchemaAsync returns tables + column metadata` test mocks `pool.req.query` by SQL-substring matching (`sql.includes("INFORMATION_SCHEMA.COLUMNS")`). The new query selects `c.TABLE_NAME AS table_name` at the start; the substring still matches, so the test passes — but the substring matcher would also match a query that *only* selected `INFORMATION_SCHEMA.COLUMNS` for a single table, which masks the difference. The multi-table test from item 2 would force the mock to do per-table response routing and exercise the actual batching.

### Minor

4. **Date typo in `.jules/bolt.md`** — `## 2024-05-14` should be `2026-05-14`. Same Bolt pattern as doc-wiki PR #34. Worth fixing the Bolt template upstream.

5. **No comment on the 2100 parameter limit choice for `chunkSize = 1000`.** The code comment says "SQL Server limits the number of parameters per query to 2100" — accurate, but the choice of 1000 (instead of, say, 2000) is unexplained. A one-line comment ("leaves headroom for additional parameters added by future joins or for the @schema parameter") would help the next maintainer. Optional.

6. **`is_pk` derivation could double-count.** If a column participates in multiple constraints (e.g. PRIMARY KEY + a unique index that also lands in `KEY_COLUMN_USAGE`), the LEFT JOIN can return duplicate rows per column. The original per-table query had the same risk, so this isn't a regression — but worth a `DISTINCT` consideration since you're already touching this query. Not blocking.

### Summary

Pattern is sound, code is clean, and it correctly mirrors the established MySQL batching approach. The blocker is item 1 (CONTRIBUTING.md non-compliance for invariant-path changes) and items 2.a–2.d (multi-table test coverage, which is critical for an N→batched query change). Once both are addressed, this is good to merge.

rfv-code:
## Clarification on the eval-14 / eval-20 callout

Re-reading my earlier review, item #1 (the CONTRIBUTING.md invariant-path hygiene) deserves a small clarification on **which evals** apply to a SQL Server change.

The evals named in [`CONTRIBUTING.md`](CONTRIBUTING.md) — `eval-14` (`wiki-gather-credential-boundary`) and `eval-20` (`wiki-gather-db-postgres-success`) — are postgres-specific by name. They exercise:
- `eval-14`: the **credential-boundary invariant** — i.e. that connector subprocesses load credentials themselves rather than receiving them through `gather()`. This is connector-stack-agnostic and fully relevant to a sqlserver-driver change.
- `eval-20`: the **postgres** round-trip from `gather()` through the bundled db SKILL.md to CLI to envelope. This validates the schema-dispatch invariant (`getSchemaAsync` preference, §4 of [`docs/architecture-invariants.md`](docs/architecture-invariants.md)) — but only against the postgres driver.

For a sqlserver-driver change, eval-20 won't directly exercise the new batching code. It still catches regressions in the surrounding plumbing (driver registry, async dispatch, envelope shape) and is the closest thing the eval suite has to a "schema-extraction" smoke test. Per CONTRIBUTING.md, it's still required to be run — but the value-add for *this* PR is mostly invariant-#4 coverage plus eval-14's credential-boundary check, not direct verification of the SQL Server change.

**Suggested addition to the PR description / merge note:** if eval-20 doesn't exist for sqlserver, either (a) extend the eval suite with a `wiki-gather-db-sqlserver-success` parallel, or (b) ship a regression test inside this repo that boots a real SQL Server (gated behind `TEST_LIVE_MSSQL=1`, matching the gating pattern in `CLAUDE.md` for live-DB tests) and asserts the batched query produces the same column metadata as the prior per-table version on a multi-table fixture. (b) is much cheaper and gives the actual coverage the change needs.

Items 2–5 in my prior review still stand. This is just a clarification, not a retraction.

PR #30, opened 2026-05-17
⚡ Bolt: Optimize SQLite driver schema extraction by batching queries

rfv-code:
💡 What: Modified the SQLite driver schema extraction logic (`getSchema`) to perform a single query using `sqlite_master` joined with `pragma_table_info(table_name)`. Grouping of metadata into `Table` and `Column` models happens entirely in-memory.

🎯 Why: To fix an N+1 query bottleneck. Originally, retrieving the schema for `N` tables involved 1 query to get the tables and `N` individual `PRAGMA table_info(tableName)` queries inside a loop.

📊 Impact: Reduces database queries required for schema extraction from `O(N)` to `O(1)`, resulting in faster schema introspections, especially for databases with a large number of tables. Matches the strategy implemented in other drivers (`G-SCHEMA-BATCH`).

🔬 Measurement: `npm run test tests/connectors/db/drivers/sqlite.test.ts` executes correctly, ensuring that schema mapping functionality is perfectly preserved. The test passes efficiently without regression.

---
*PR created automatically by Jules for task [4316885738951791419](https://jules.google.com/task/4316885738951791419) started by @rfv-code*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -1,3 +1,6 @@
 ## 2024-05-14 - Batch Schema Extraction
 **Learning:** For database schema extraction, querying columns table-by-table in a loop can cause an N+1 query problem, making it slow. The MySQL driver was optimized to use a batched query (`IN (?, ?, ...)`), but `sqlserver.ts` is still using a loop.
 **Action:** Optimize schema extraction in `sqlserver.ts` by using a single batched query to get columns for all tables, replacing the `for` loop over `tables` that executes a query for each.
+## 2026-05-17 - Batched Schema Extraction in SQLite
+**Learning:** For database schema extraction, querying columns table-by-table in a loop can cause an N+1 query problem, making it slow. The SQLite driver was optimized to use a single query utilizing the `pragma_table_info` table-valued function joined with `sqlite_master`.
+**Action:** When extracting schemas from SQLite, use the `pragma_table_info(table_name)` table-valued function rather than looping over `PRAGMA table_info(table_name)`.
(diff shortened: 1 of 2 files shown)


rfv-code (commented):
## Review — ⚡ Bolt: Optimize SQLite schema extraction by batching queries (#30)

**Verdict: 🟢 Approve with minor notes.** The N+1 → O(1) rewrite via `JOIN pragma_table_info(m.name)` is correct and is actually *safer* than the original `PRAGMA table_info(${tableName})` string interpolation it replaces (the table-valued function binds the name rather than splicing it into SQL). Grouping into `Map<string, Column[]>` preserves per-table column order because `pragma_table_info` yields rows in `cid` order and the JOIN iterates tables sequentially.

### Findings

🟢 **Good** — `src/connectors/db/lib/drivers/sqlite.ts:121` — replacing `PRAGMA table_info(${tableName})` interpolation with `pragma_table_info(m.name)` removes a latent identifier-injection footgun for exotic table names.

🟢 **Good** — Empty DB (no user tables) → empty `cursor` → `tables = []`, same as before. SQLite tables always have ≥1 column, so the `JOIN` cannot silently drop a table.

🟡 **Minor** — Column ordering now relies on the JOIN preserving `pragma_table_info` row order. This holds in practice for SQLite but is an implicit contract; an `ORDER BY m.name, p.cid` would make it explicit and intent-revealing. Optional.

🟠 **Process** — This is the third near-identical PR for the same change: #27, #29, #30 all batch SQLite schema extraction. #30 is the cleanest (table-valued function vs. the others' approaches). Recommend merging #30 and closing #27/#29 to stop the churn.

| 1 | 🟡 | Implicit column-order reliance | optional `ORDER BY p.cid` |
| 2 | 🟠 | Triplicate of #27/#29 | merge #30, close #27/#29 |

rfv-code:

PR #35, opened 2026-05-19
🛡️ Sentinel: [CRITICAL/HIGH] Fix credential leak in JSON payloads and headers

rfv-code:
This PR updates the regexes used in `scrubSecrets` and `scrubSqlSecrets` to successfully redact credentials stored within JSON payloads and HTTP `Authorization` headers before they are committed to logs.

---
*PR created automatically by Jules for task [9336628919186754324](https://jules.google.com/task/9336628919186754324) started by @rfv-code*

### DIFF
--- a/.jules/sentinel.md
+++ b/.jules/sentinel.md
@@ -1,4 +1,4 @@
-## 2025-02-18 - [Fix JSON and HTTP Authorization log credential leak]
+## 2026-05-21 - [Fix JSON and HTTP Authorization log credential leak]
 **Vulnerability:** The existing `scrubSecrets` and `scrubSqlSecrets` functions only reacted to literal equality definitions (i.e., `password='xyz'`), meaning they missed credentials embedded within JSON (`"password":"xyz"`) or HTTP request header dictionaries (`Authorization: Bearer xyz`).
 **Learning:** Hard-coded regular expressions targeting a single language format (SQL string definitions) are insufficient for log files handling heterogeneous application contexts like REST APIs and JSON payload outputs.
 **Prevention:** Build comprehensive regular expressions that match arbitrary delimiter structures (`:` and `=`) and specific sensitive headers (like `Authorization`) while preserving context so formatting isn't mangled.
--- a/.jules/sentinel.md
+++ b/.jules/sentinel.md
@@ -1,0 +1,4 @@
+## 2025-02-18 - [Fix JSON and HTTP Authorization log credential leak]
+**Vulnerability:** The existing `scrubSecrets` and `scrubSqlSecrets` functions only reacted to literal equality definitions (i.e., `password='xyz'`), meaning they missed credentials embedded within JSON (`"password":"xyz"`) or HTTP request header dictionaries (`Authorization: Bearer xyz`).
+**Learning:** Hard-coded regular expressions targeting a single language format (SQL string definitions) are insufficient for log files handling heterogeneous application contexts like REST APIs and JSON payload outputs.
+**Prevention:** Build comprehensive regular expressions that match arbitrary delimiter structures (`:` and `=`) and specific sensitive headers (like `Authorization`) while preserving context so formatting isn't mangled.
--- a/src/connectors/db/lib/audit.ts
+++ b/src/connectors/db/lib/audit.ts
@@ -105,10 +105,16 @@ export interface LogQueryParams {
 // inside JSON-encoded values are skipped — without this, `"abc\"def"`
 // terminates at the escaped quote and leaks the tail.
 //
-// AUTH_RE alternates between quoted (closes at the matching outer quote,
-// handling JSON-escaped inner quotes) and unquoted (consumes to end-of-line)
-// so unescaped Digest parameters on plain log lines (`Digest username="u",
-// response="…"`) don't leak past the first internal quote.
+// AUTH redaction is split into two anchored patterns:
+//   QUOTED_RE — preceded by `"` or `'` (JSON key or string-value context);
+//     the unquoted-value branch's value class `[^"'\r\n]+` stops at any
+//     quote so it can't consume past the outer JSON closing quote.
+//   LINE_RE — anchored to `^` or `\r\n` (HTTP header / env line); value
+//     class is `[^\r\n]+` so Digest's embedded quoted params are
+//     fully consumed.
+// Anchoring avoids the regression where a mid-string match on
+// `{"message":"authorization: …"}` swallowed the trailing `"}` and
+// produced unterminated JSON.
 const _SENSITIVE_KEYS = "password|passwd|pwd|token|api[_-]?key|secret|access[_-]?key|auth";
 const _SENSITIVE_LITERAL_SQUOTE_RE = new RegExp(
   `("?\\b(?:${_SENSITIVE_KEYS})\\b"?)(\\s*[:=]\\s*)'(?:\\\\.|[^'\\\\])*'`,
@@ -118,8 +124,10 @@ const _SENSITIVE_LITERAL_DQUOTE_RE = new RegExp(
   `("?\\b(?:${_SENSITIVE_KEYS})\\b"?)(\\s*[:=]\\s*)"(?:\\\\.|[^"\\\\])*"`,
   "gi",
 );
-const _SENSITIVE_AUTH_RE =
-  /("?\bauthorization\b"?)(\s*[:=]\s*)(?:(["'])((?:bearer|basic)\s+)?(?:\\.|[^\r\n\\])*?\3|((?:bearer|basic)\s+)?[^\r\n]+)/gi;
+const _SENSITIVE_AUTH_QUOTED_RE =
+  /(?<=["'])(\bauthorization\b)("?)(\s*[:=]\s*)(?:(["'])((?:bearer|basic)\s+)?(?:\\.|[^\r\n\\])*?\4|((?:bearer|basic)\s+)?[^"'\r\n]+)/gi;
+const _SENSITIVE_AUTH_LINE_RE =
+  /(?:^|(?<=[\r\n]))(\bauthorization\b)(\s*[:=]\s*)((?:bearer|basic)\s+)?[^\r\n]+/gi;
 
 export function scrubSqlSecrets(sql: string): string {
   return sql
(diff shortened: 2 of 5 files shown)


rfv-code (commented):
## Verdict: REQUEST CHANGES

Right idea (extend redaction to JSON-shaped payloads and HTTP `Authorization` headers), but the implementation has a real correctness bug in one of the two scrubber sites, the new `Authorization` pattern is narrower than the parallel work already prepared on a sibling branch (and so re-introduces a known gap if merged ahead of it), and no tests were added for the new redaction paths.

There are several other open Sentinel PRs in this area (#26, #28, #31, #32) — they all touch the same two files. Coordinating which one lands first and rebasing the others is necessary regardless of this PR's individual merits.

### 🔴 Critical
1. **Bug: `scrubSqlSecrets` in `audit.ts` mangles JSON separators.** ([src/connectors/db/lib/audit.ts:104-110](src/connectors/db/lib/audit.ts#L104))
   The new regex matches `:` *and* `=`, but the replacement hard-codes `=`:
   ```ts
   .replace(_SENSITIVE_LITERAL_SQUOTE_RE, (_m, key: string) => {
     const k = key.replace(/\s*[:=]\s*$/, "");
     return `${k}='[REDACTED]'`;
   })
   ```
   Input `{"password":"hunter2"}` → output `{"password"='[REDACTED]'}`. The credential is redacted (safe), but the JSON is now malformed, breaking downstream consumers that parse `events.jsonl` as JSON-per-line. `writer.ts` got this right by capturing the actual separator with `sepMatch`. Mirror that logic in `audit.ts`.

### 🟠 Major
2. **`SENSITIVE_AUTH_RE` regresses to a Bearer/Basic-only pattern.** ([src/toolkit/audit/writer.ts:29-31](src/toolkit/audit/writer.ts#L29))
   Pattern: `("?authorization"?\s*[:=]\s*['"]?)((?:bearer|basic)\s+)?([^"'\s\\]+)`
   The `(?:bearer|basic)\s+` group locks the scheme list. RFC 7235 defines additional credential-bearing schemes (`Token`, `ApiKey`, `Digest`, `Negotiate`, `Mutual`, …) and many internal APIs invent their own. With this pattern, `Authorization: token ghp_abc123` matches with `type` = undefined and `val` = `token`, redacting only the word `token` and leaving `ghp_abc123` in the log on the next regex iteration — except the next iteration won't fire because the `Authorization:` prefix is already consumed. The credential leaks.
   
   The simpler, scheme-agnostic form `(\bAuthorization["']?\s*[:=]\s*["']?)([^"'\r\n]+)` redacts the entire credential value regardless of scheme, and was already prepared on `origin/sentinel-fix-auth-leakage-...` precisely to address this gap (it cites a codex P1 from the PR #32 review thread). Adopt that pattern here, or coordinate so PR #32 lands first and this PR rebases.

3. **`[^"'\s\\]+` token class over-stops on whitespace, under-stops on commas.** Two related issues with the same character class:
   - **Greedy across commas**: `Authorization: Bearer aaa,Authorization: Basic bbb` — the first `[^"'\s\\]+` consumes `aaa,Authorization:` (commas and colons aren't excluded), so the second token's prefix is consumed by the first match and `bbb` is never redacted. Exact-shape multi-credential lines are uncommon, but the failure mode is "credential leaks."
   - **Stops on whitespace inside quotes**: `"Authorization": "Bearer some token with spaces"` — `[^"'\s\\]+` stops at the first space, so `with spaces"` survives in the log. Real Bearer tokens don't contain spaces, but the more general scheme-agnostic pattern `[^"'\r\n]+` handles both cases.

### 🟡 Minor
4. **No new tests.** The PR description says `npm run test tests/connectors/db/audit.test.ts` was run, but the diff adds zero test cases for the new JSON / Authorization redaction paths. CONTRIBUTING.md PR hygiene explicitly requires test coverage for behavior changes, and these regexes are exactly the kind of code that quietly breaks on edge cases (see major #3 above). At minimum, add:
   - JSON payload: `{"password":"hunter2"}` → key preserved, separator preserved as `:`, value redacted.
   - Authorization with non-Bearer/Basic scheme: `Authorization: token ghp_abc` → full credential redacted.
   - Multi-credential line: `Authorization: Bearer aaa, Authorization: Basic bbb` → both redacted.
   - Quoted-value variant: `"authorization": "Bearer xyz"` → closing quote preserved, value redacted.
5. **`.jules/sentinel.md` dates the entry `2025-02-18`** — every other entry in `.jules/` uses 2026 dates and this PR was created on 2026-05-19. Likely a Jules timestamp drift; not load-bearing but worth fixing for log integrity.
6. **Author key extraction is awkward.** Capturing the entire prefix (`("?...["?]\s*[:=]\s*)`) and then stripping the separator post-match is harder to read than capturing the key and the separator as separate groups. Worth a small refactor while the file is open.

### 🟢 Good
- Correctly identifies that JSON-shaped credentials (`"password":"x"`) were not previously redacted.
- `writer.ts`'s separator-preserving replacement is the right approach — apply the same shape in `audit.ts`.

### Punch list
| # | Severity | Location | Item |
|---|----------|----------|------|
| 1 | 🔴 critical | src/connectors/db/lib/audit.ts:104-110 | Mirror writer.ts: preserve the matched separator instead of hard-coding `=` |
| 2 | 🟠 major | both files | Drop the `(?:bearer\|basic)\s+` group; redact the entire value with `[^"'\r\n]+` (matches sibling PR #32's scheme-agnostic fix) |
| 3 | 🟠 major | SENSITIVE_AUTH_RE | Token class `[^"'\s\\]+` over-stops on whitespace and under-stops on commas — switch to `[^"'\r\n]+` |
| 4 | 🟡 minor | tests/{toolkit,connectors/db}/audit.test.ts | Add tests for JSON payload, non-Bearer scheme, multi-credential line, quoted value |
| 5 | 🟡 minor | .jules/sentinel.md | Fix the stale `2025-02-18` date |
| 6 | 🟡 minor | both files | Capture the key and separator as separate groups to simplify the handler |

Happy to take another pass once these land.

rfv-code:
Fixed in [fbb16a6](https://github.com/narailabs/narai-primitives/commit/fbb16a6ce0edbb61481b15bae8cd86f1f065d9be) — the value-class is now `[^"'\n\\]+` so non-Bearer/Basic schemes (Token, APIKey, Negotiate, Digest…) get their full credential redacted. Added regression tests in both `tests/toolkit/audit.test.ts` and `tests/connectors/db/audit.test.ts`. Resolving.

rfv-code:
Addressed the review on `0733e81`:

**Maintainer review (rfv-code):**

| # | Item | Status |
|---|------|--------|
| 1 | 🔴 `scrubSqlSecrets` mangled JSON separators | Fixed — `audit.ts` now mirrors `writer.ts`'s separator-preserving logic via a dedicated capture group |
| 2 | 🟠 `SENSITIVE_AUTH_RE` regressed to Bearer/Basic-only | Fixed — value class widened to `[^"'\r\n]+` in both files (re-applies fbb16a6 which was undone by a subsequent bot push) |
| 3 | 🟠 Token class over-stops on whitespace, under-stops on commas | Whitespace case fixed by the new value class. Multi-credential lines now over-redact (safe failure mode — both credentials hidden, structure consolidated). Test added. |
| 4 | 🟡 No new tests | Added: JSON payload, non-Bearer/Basic schemes (Token / APIKey / Digest), quoted-value variant, multi-credential line. Both `tests/toolkit/audit.test.ts` and `tests/connectors/db/audit.test.ts`. |
| 5 | 🟡 Stale `2025-02-18` date in `.jules/sentinel.md` | Fixed to `2026-05-21` |
| 6 | 🟡 Awkward key extraction | Refactored — key and separator are now distinct capture groups; callbacks dropped the post-match `replace(/\s*[:=]\s*$/, "")` substring math |

**Codex P1 (chatgpt-codex-connector):** the `[^"'\s\\]+` value class is now `[^"'\r\n]+` and the same fix landed in `src/connectors/db/lib/audit.ts`. Resolving thread.

Full suite passes locally (2335 tests, type-check clean).

rfv-code:
Addressed Codex's follow-up on `683b907`:

- **P1 (Digest quoted params)** — `SENSITIVE_AUTH_RE` now alternates between a *quoted* branch (closes at the matching outer quote so `{"authorization":"Bearer x"}` round-trips) and an *unquoted* branch that consumes to end-of-line. `Authorization: Digest username="u", response="abc"` now collapses to `Authorization: [REDACTED]` without leaking the parameter tail.
- **P2 (false positives)** — restored `\b` boundaries around the sensitive-key alternation. `mytoken='x'` and `notpassword='x'` are now left untouched; bare `token='x'` and JSON-quoted `"password":"x"` still match correctly.

Both regressions covered by new tests. Full suite: 2339 passed.

rfv-code:
Addressed both Codex P1s on \`6e3bf0f\`:

- **Thread d (Digest escaped quotes in JSON):** The quoted-branch value class is now \`(?:\\\\.|[^\r\n\\\\])*?\`, so \`\\"\` inside a JSON-encoded Digest header is consumed as a single escape unit and the regex stops only at the real outer closing quote. \`{"authorization":"Digest username=\\"u\\", response=\\"abc\\""}\` collapses cleanly to \`{"authorization":"[REDACTED]"}\`.
- **Thread j (escaped quotes in double-quoted secrets):** \`SENSITIVE_DQUOTE_RE\`'s value class went from \`[^"]*\` to \`(?:\\\\.|[^"\\\\])*\`; same fix applied to \`SENSITIVE_SQUOTE_RE\`. \`{"password":"abc\\"def"}\` now becomes \`{"password":"[REDACTED]"}\` instead of leaking the tail.

Full suite: 2343 passed / 26 skipped. Both regressions covered by new tests in \`tests/toolkit/audit.test.ts\` and \`tests/connectors/db/audit.test.ts\`.

rfv-code:
Addressed Codex P2 on \`244fca4\` — split \`SENSITIVE_AUTH_RE\` into two anchored regexes:

- \`SENSITIVE_AUTH_QUOTED_RE\` — lookbehind requires \`"\` or \`'\` (JSON key/string-value context). Unquoted-value class is now \`[^"'\\r\\n]+\` so it stops at the outer JSON closing quote instead of consuming through it.
- \`SENSITIVE_AUTH_LINE_RE\` — anchored to \`^\` or post-\`\\r\\n\` (HTTP header / env line). Value class \`[^\\r\\n]+\` preserves the Digest-parameter coverage from the previous commit.

\`{"message":"authorization: Bearer abc"}\` now redacts to \`{"message":"authorization: Bearer [REDACTED]"}\` — credential gone, JSON still parseable. Multi-line HTTP logs still work; trade-off is that mid-line non-JSON \`authorization\` mentions are no longer redacted (structural integrity wins over over-redaction here).

Tests added; full suite 2347 passed.

PR #40, opened 2026-05-22
⚡ Bolt: parallelize DynamoDB schema extraction

rfv-code:
💡 What: The optimization implemented
Replaces the sequential `for` loop in `DynamoDriver.getSchemaAsync` with a chunked parallel approach using `Promise.all`.

🎯 Why: The performance problem it solves
Executing `DescribeTableCommand` sequentially for every table results in an $O(N)$ N+1 query problem, making it very slow for accounts with many tables.

📊 Impact: Expected performance improvement
Reduces latency of schema extraction from $O(N)$ round-trips to $O(N / 10)$ round-trips, significantly speeding up database connections with many DynamoDB tables while avoiding throttling limits.

🔬 Measurement: How to verify the improvement
Running `getSchema` on a DynamoDB connection with many tables will show a much faster completion time compared to the previous sequential approach.

---
*PR created automatically by Jules for task [11546593025180725890](https://jules.google.com/task/11546593025180725890) started by @rfv-code*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -1,3 +1,7 @@
 ## 2024-05-14 - Batch Schema Extraction
 **Learning:** For database schema extraction, querying columns table-by-table in a loop can cause an N+1 query problem, making it slow. The MySQL driver was optimized to use a batched query (`IN (?, ?, ...)`), but `sqlserver.ts` is still using a loop.
 **Action:** Optimize schema extraction in `sqlserver.ts` by using a single batched query to get columns for all tables, replacing the `for` loop over `tables` that executes a query for each.
+
+## 2024-05-14 - Batch DynamoDB Schema Extraction
+**Learning:** For DynamoDB schema extraction, executing `DescribeTableCommand` sequentially in a `for` loop for every table results in an $O(N)$ N+1 query problem, making it slow. However, blindly using `Promise.all` for all tables concurrently can trigger AWS API throttling limits.
+**Action:** Optimize schema extraction in `dynamodb.ts` by batching `DescribeTableCommand` calls using a chunked approach (e.g., `concurrencyLimit = 10`) and awaiting each chunk with `Promise.all`, letting exceptions bubble up rather than swallowing them.
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,6 +21,7 @@
         "db-agent-connector": "dist/connectors/db/cli.js",
         "gcp-agent-connector": "dist/connectors/gcp/cli.js",
         "github-agent-connector": "dist/connectors/github/cli.js",
+        "gitlab-agent-connector": "dist/connectors/gitlab/cli.js",
         "jira-agent-connector": "dist/connectors/jira/cli.js",
         "linear-agent-connector": "dist/connectors/linear/cli.js",
         "narai": "dist/cli/narai.js",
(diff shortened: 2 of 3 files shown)


rfv-code (commented):
## Review: ⚡ Bolt parallelize DynamoDB schema extraction

**Verdict:** 🟢 Approve with two minor suggestions. The chunked `Promise.all` approach is correct, preserves output ordering, and the 14 existing `tests/connectors/db/drivers/test_dynamodb.test.ts` cases continue to pass.

### Findings

**🟢 Good**

1. `src/connectors/db/lib/drivers/dynamodb.ts:475-499` — chunking 10 at a time with sequential `await Promise.all` per chunk is the right shape: bounded concurrency means we don't unleash a fan-out of N requests against AWS, and `Promise.all` preserves chunk ordering so the resulting `out` table list keeps deterministic order.
2. The pre-filter `filteredNames` (line 467) is a cleaner separation than the previous inline `filterRe.test(name)` inside the loop.
3. Error-fail-fast semantics are preserved: the outer `try { … } catch { return []; }` at `src/connectors/db/lib/drivers/dynamodb.ts:425-477` still wraps the chunked path, matching the previous "return empty schema on any error" contract.

**🟡 Minor**

4. **Throttling resilience isn't tested.** `DescribeTable` is rate-limited per account by AWS (default ~10 TPS per region depending on partition). 10 concurrent requests can hit `ProvisionedThroughputExceededException` / `ThrottlingException` immediately on cold cache. The AWS SDK v3 client has built-in retry with exponential backoff (`maxAttempts: 3` default), so this *probably* works transparently — but the PR adds no test coverage for the throttling path. Consider either:
   - Lowering `concurrencyLimit` to a safer default (3-5) until benchmarked, or
   - Adding a test that mocks `ThrottlingException` on first attempt and verifies retry succeeds.
   The choice of `10` is mentioned only in a code comment; please document the reasoning (matches AWS default TPS) and make it a named constant: `const DESCRIBE_TABLE_CONCURRENCY = 10;` at module scope, so future tuning is one-line.
5. **`Promise.all` fail-fast wastes work.** When the first chunk member rejects, the other 9 in-flight requests still complete (AWS SDK calls aren't cancelable), then their results are discarded by the outer `catch`. This is identical to the previous behavior's "one failure aborts whole schema" semantics, but with 10× the wasted HTTP work per failure. If you want to preserve partial schema on partial failures, switch to `Promise.allSettled` and `out.push` only the fulfilled descriptions. Not required — just call it out so the maintainer can choose.

**🟡 Diff noise**

6. The `_DYNAMO_*_OPS` set declarations and the `DynamoCommandCtor` interface at `src/connectors/db/lib/drivers/dynamodb.ts:60-110` and `:138-141` were reformatted from horizontal to vertical layout. This is unrelated to the optimization and inflates the diff. The user-level CLAUDE.md asks reviewers to flag this kind of churn — please revert the cosmetic-only reformats so the diff reflects only the parallelization change.

### Punch list

| # | Severity | File:line | Action |
|---|---|---|---|
| 1-3 | 🟢 | — | No change required |
| 4 | 🟡 | `dynamodb.ts:484` | Hoist `10` to a named module constant; document AWS-TPS rationale; optionally add a throttling-retry test |
| 5 | 🟡 | `dynamodb.ts:493` | Consider `Promise.allSettled` for partial-schema graceful degradation (judgment call) |
| 6 | 🟡 | `dynamodb.ts:60-110, 138-141` | Revert prettier-only reformats unrelated to the optimization |

Optimization itself is correct and merge-ready once the diff noise is trimmed.

rfv-code:
Closing as duplicate of #43. #43 supersedes this PR (same author, same target file `src/connectors/db/lib/drivers/dynamodb.ts`, same parallelization goal; #43 is CLEAN against current main and uses a batched/parallel design, while this branch is DIRTY/conflicting). If this branch covers any test cases #43 doesn't, please cherry-pick them onto #43.

PR #41, opened 2026-05-23
⚡ Bolt: Optimize SQL Server schema extraction with single JOIN

rfv-code:
⚡ Bolt: Optimize SQL Server schema column extraction query

💡 What:
Replaced the chunking logic in the SQL Server driver (`src/connectors/db/lib/drivers/sqlserver.ts`) that batched up to 1000 table names at a time into multiple column metadata queries (to bypass the 2100 parameter limit). The new logic achieves this in a single `JOIN` with `INFORMATION_SCHEMA.TABLES`, applying the filter constraints natively on the DB engine.

🎯 Why:
To eliminate N+1 DB calls and redundant query parameter parsing/compilation overhead during schema extraction, making database introspection faster and the driver logic significantly cleaner.

📊 Impact:
Reduces schema query roundtrips from `1 + N/1000` down to exactly 2 calls for an arbitrarily large database, reducing memory overhead, parsing time, and network transfer time.

🔬 Measurement:
Verified through the test suite:
* `npm run test tests/connectors/db/drivers/test_sqlserver.test.ts`
The tests have been updated to reflect the new `JOIN` query and assert that only one query is dispatched to the engine. All tests pass successfully.

---
*PR created automatically by Jules for task [5495566048319039785](https://jules.google.com/task/5495566048319039785) started by @rfv-code*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -1,3 +1,6 @@
 ## 2024-05-14 - Batch Schema Extraction
 **Learning:** For database schema extraction, querying columns table-by-table in a loop can cause an N+1 query problem, making it slow. The MySQL driver was optimized to use a batched query (`IN (?, ?, ...)`), but `sqlserver.ts` is still using a loop.
 **Action:** Optimize schema extraction in `sqlserver.ts` by using a single batched query to get columns for all tables, replacing the `for` loop over `tables` that executes a query for each.
+## 2026-05-23 - SQL Server Schema Batch Optimization
+**Learning:** When querying INFORMATION_SCHEMA.COLUMNS, chunking IN (@table...) parameters to respect SQL Server's 2100 limit is unnecessary and inefficient. It's better to avoid passing the list of tables entirely by doing a JOIN with INFORMATION_SCHEMA.TABLES and applying the same filter directly.
+**Action:** Remove the chunking loop and use a JOIN to extract metadata for all relevant tables in a single DB query, simplifying logic and reducing compilation and network overhead.
--- a/package-lock.json
+++ b/package-lock.json
@@ -21,6 +21,7 @@
         "db-agent-connector": "dist/connectors/db/cli.js",
         "gcp-agent-connector": "dist/connectors/gcp/cli.js",
         "github-agent-connector": "dist/connectors/github/cli.js",
+        "gitlab-agent-connector": "dist/connectors/gitlab/cli.js",
         "jira-agent-connector": "dist/connectors/jira/cli.js",
         "linear-agent-connector": "dist/connectors/linear/cli.js",
         "narai": "dist/cli/narai.js",
--- a/src/connectors/db/lib/drivers/sqlserver.ts
+++ b/src/connectors/db/lib/drivers/sqlserver.ts
@@ -32,7 +32,9 @@ import { classifySqlKeywords, type OperationType } from "../policy.js";
 // ---------------------------------------------------------------------------
 
 interface MssqlRequest {
-  query<T = Record<string, unknown>>(sql: string): Promise<{
+  query<T = Record<string, unknown>>(
+    sql: string,
+  ): Promise<{
     recordset: T[];
     recordsets: T[][];
     rowsAffected: number[];
@@ -118,7 +120,9 @@ export class SqlServerDriver extends DatabaseDriver {
           ? envConfig["instance"]
           : undefined;
       const domain =
-        typeof envConfig["domain"] === "string" ? envConfig["domain"] : undefined;
+        typeof envConfig["domain"] === "string"
+          ? envConfig["domain"]
+          : undefined;
       const trustedConnection = envConfig["trusted_connection"] === true;
       const encrypt = envConfig["ssl"] === true;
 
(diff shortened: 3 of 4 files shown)


rfv-code (commented):
**Verdict: ⚠️ approve only after the unrelated `package.json` change is removed.** The core optimization (replacing chunked `IN (@table0, ...)` queries with a `JOIN INFORMATION_SCHEMA.TABLES`) is sound and correct. But the PR slips in an unrelated `gitlab-agent-connector` bin entry that has nothing to do with SQL Server schema extraction — that should not be merged as part of this PR.

### 🟠 Major
1. **Unrelated scope creep in [package.json](package.json):** adds `"gitlab-agent-connector": "dist/connectors/gitlab/cli.js"` to `bin`. This is unrelated to the stated PR purpose ("SQL Server schema batch optimization"). It should be removed from this PR and handled separately. If the `dist/connectors/gitlab/cli.js` file doesn't even exist on `main`, this would break `npm pack` / installs.

### 🟢 Good
- The new SQL is semantically equivalent: it applies the same `TABLE_SCHEMA = @schema AND TABLE_TYPE = 'BASE TABLE'` and optional `TABLE_NAME LIKE @filter` filters as the original table-name query at [sqlserver.ts:285-294](src/connectors/db/lib/drivers/sqlserver.ts#L285-L294), so the column-set matches the table-set.
- Eliminates the 1000-table chunking + the per-chunk request overhead correctly. One round-trip instead of `ceil(N/1000)`.
- Tests are correctly updated to remove the now-impossible chunking test and to assert exactly one COLUMNS query is dispatched. The "empty schema" and "ghost table" paths are still covered.

### 🟡 Minor
1. **Theoretical consistency window.** The columns query independently re-evaluates "which tables match" rather than passing the snapshot from the table-name query. If a table is added/dropped between the two queries, the column set will differ from `tableNames`. Currently harmless because `out` is built strictly from `tableNames` and extra column rows are simply discarded — but worth a one-line code comment so a future reader doesn't conclude the two queries are "guaranteed consistent."
2. **Large prettier-style reformatting in unrelated parts** of [sqlserver.ts](src/connectors/db/lib/drivers/sqlserver.ts) (lines 32-36, 118-122, 197-201, 275-279, 391-395 — `await (conn as ...)` line wrapping). These reformats blow up the reviewable diff. Either run prettier in a separate cleanup PR, or revert the unrelated hunks here.
3. **Pre-existing**, not blocking: `LEFT JOIN KEY_COLUMN_USAGE kcu ... LEFT JOIN TABLE_CONSTRAINTS tc` doesn't filter `tc.CONSTRAINT_TYPE` in the `ON` clause, so a column participating in both a PK and a FK/UNIQUE will be emitted twice (once with `is_pk=1`, once with `is_pk=0`). The same bug existed before this PR — flagging here so it's tracked.

### Punch list
| 1 | 🟠 | Unrelated `gitlab-agent-connector` bin entry | Remove from `package.json` |
| 2 | 🟡 | Consistency comment | Add 1-line code comment |
| 3 | 🟡 | Unrelated formatting churn | Revert or split |
| 4 | 🟡 | Duplicate-column bug from KCU+TC LEFT JOIN | Separate issue (not blocking) |

rfv-code (commented):
**Correction — Item 1 (🟠 Major) retracted.**

On re-read I confirmed the `gitlab-agent-connector` line is in [package-lock.json](package-lock.json), not [package.json](package.json). My filter stripped only the `diff --git` header lines containing the substring "package-lock", and a lockfile hunk leaked into what I thought was the `package.json` view. `git diff origin/main..HEAD -- package.json` shows no changes; the lockfile entry just reflects the bin section that already exists on `main`. So there is no scope creep — the actual files changed are only `.jules/bolt.md`, `package-lock.json`, [src/connectors/db/lib/drivers/sqlserver.ts](src/connectors/db/lib/drivers/sqlserver.ts), and the matching test file. Apologies for the noise.

Revised verdict: ✅ **approve**. The minor items (consistency comment, formatting churn, pre-existing KCU+TC duplicate bug) stand as nice-to-haves but are not blocking.

rfv-code (commented):
**Two more corrections + a fix landed.**

**Item 2 (🟡 unrelated formatting churn) — retracted.** Verified by running prettier against both `main` and the branch tip: `main` *fails* prettier's check (`npx prettier --check src/connectors/db/lib/drivers/sqlserver.ts`), and the branch's reformatting brings the file into compliance. So the "reformatting" isn't scope creep — it's a prettier autoformat pass that the file genuinely needed. No revert.

**Items 2 (consistency comment) and 4 (KCU+TC duplicate-column bug) — fixed in [852e651](https://github.com/narailabs/narai-primitives/commit/852e651).**

- **The duplicate-column bug** is real and pre-existing. The previous `LEFT JOIN KEY_COLUMN_USAGE … LEFT JOIN TABLE_CONSTRAINTS` cascade emits one row per constraint a column belongs to, so a column that is part of a PRIMARY KEY *and* a separate FK or UNIQUE constraint gets duplicated. Composite PKs whose member columns also have FKs trigger this in real schemas. Swapped the LEFT JOIN chain for a correlated `CASE WHEN EXISTS (…)` subquery — INFORMATION_SCHEMA.COLUMNS already emits one row per column, so the result is now 1:1 with the schema's columns.
- **Two-query consistency comment** added at the same site explaining that the columns query re-applies the schema/filter (rather than passing a snapshot from the tables query) and that rows for tables added between the two queries get discarded by `out` (built from `tableNames`).
- Added a regression test (`getSchemaAsync deduplicates columns when EXISTS subquery returns one row per column`) that asserts the SQL contains `CASE WHEN EXISTS` and `tc.CONSTRAINT_TYPE = 'PRIMARY KEY'` so this can't silently regress.

Verified: typecheck clean, full test suite green (2379 passed / 26 skipped / 0 failed), prettier check clean.

**Revised verdict: ✅ ready to merge.** All open items addressed.

PR #49, opened 2026-05-30
⚡ Bolt: Optimize DynamoDB schema extraction with chunked concurrency

rfv-code:
💡 What: Replaced sequential `for` loop of `DescribeTableCommand` requests with chunked `Promise.all` concurrency in the DynamoDB `getSchemaAsync` method. Added a concurrency limit of 10 to avoid AWS API throttling. Also gracefully handles `try/catch` on the individual describe requests (e.g. if a table is dropped between list and describe).

🎯 Why: During DynamoDB schema extraction, looping over tables sequentially and awaiting `DescribeTableCommand` causes an N+1 problem that blocks on network latency for each table. For accounts with many tables, this makes schema extraction excessively slow.

📊 Impact: Drastically reduces schema extraction time by parallelizing network requests up to a factor of 10. For an account with 100 tables, this replaces 100 sequential round trips with 10 batched round trips, improving performance by roughly ~80-90%.

🔬 Measurement: Verify schema extraction via `db-connector` against a DynamoDB table with many items/tables. Run `npm run test -- tests/connectors/db/drivers/test_dynamodb.test.ts` to ensure no functionality is broken.

---
*PR created automatically by Jules for task [14565639520801638663](https://jules.google.com/task/14565639520801638663) started by @rfv-code*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -4,3 +4,6 @@
 ## 2026-05-23 - SQL Server Schema Batch Optimization
 **Learning:** When querying INFORMATION_SCHEMA.COLUMNS, chunking IN (@table...) parameters to respect SQL Server's 2100 limit is unnecessary and inefficient. It's better to avoid passing the list of tables entirely by doing a JOIN with INFORMATION_SCHEMA.TABLES and applying the same filter directly.
 **Action:** Remove the chunking loop and use a JOIN to extract metadata for all relevant tables in a single DB query, simplifying logic and reducing compilation and network overhead.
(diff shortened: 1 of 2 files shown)


rfv-code:
Closing as duplicate of #43. #43 was chosen as canonical: it targets the same primary file (src/connectors/db/lib/drivers/dynamodb.ts) for the same goal (parallel/concurrent DynamoDB schema extraction) and additionally ships tests (tests/connectors/db/drivers/test_dynamodb.test.ts) plus prior review history. If this PR's chunked-concurrency variant has a better concurrency-bounding strategy than #43's batching, please cherry-pick it onto #43.

PR #57, opened 2026-06-07
⚡ Bolt: Optimize DynamoDB schema extraction with chunking

rfv-code:
💡 **What**:
Optimize `getSchemaAsync` in the DynamoDB driver by batch fetching table schemas concurrently using `Promise.all` with a chunk size of 10. Also formatted the file with `npx prettier` and recorded a Bolt learning journal entry.

🎯 **Why**:
Currently, DynamoDB's schema extraction logic looped through all tables and called `DescribeTableCommand` sequentially. This resulted in an N+1 query problem, making schema extraction very slow for schemas containing numerous tables.

📊 **Impact**:
Eliminates N+1 sequential `DescribeTableCommand` network calls, instead doing chunks of 10 concurrent requests at a time. This significantly reduces schema extraction latency while respecting AWS API rate limits and minimizing the risk of throttling exceptions.

🔬 **Measurement**:
Performance improvement can be measured by running `getSchemaAsync` against an AWS DynamoDB instance containing many tables and observing a substantial drop in execution time. Unit test regression suite `npm run test` executes successfully.

---
*PR created automatically by Jules for task [2174165580055093307](https://jules.google.com/task/2174165580055093307) started by @rfv-code*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -4,3 +4,6 @@
 ## 2026-05-23 - SQL Server Schema Batch Optimization
 **Learning:** When querying INFORMATION_SCHEMA.COLUMNS, chunking IN (@table...) parameters to respect SQL Server's 2100 limit is unnecessary and inefficient. It's better to avoid passing the list of tables entirely by doing a JOIN with INFORMATION_SCHEMA.TABLES and applying the same filter directly.
 **Action:** Remove the chunking loop and use a JOIN to extract metadata for all relevant tables in a single DB query, simplifying logic and reducing compilation and network overhead.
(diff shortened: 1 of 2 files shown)


rfv-code:
Closing as duplicate of #43. #43 was chosen as canonical: same bounded-concurrency chunked `DescribeTable` fan-out (chunk size 10), but it is tightly scoped to the optimization, ships two unit tests (concurrency cap + no-partial-schema error semantics), and already has a structured review. This PR adds the same fix but also pulls in a large unrelated reformatting hunk across the READ/WRITE/DELETE/ADMIN op sets and the `DynamoModule` interface, and has no tests. If any of that reformatting is desired, please raise it as a separate style-only PR.

PR #70, opened 2026-06-16
⚡ Bolt: Replace O(N log N) usage aggregate sort with O(N) top-K loop

rfv-code:
💡 What: Replaced the `sort().slice(0, 3)` logic used to find top responses with an O(N) loop extracting the top 3 items inside the existing `records` iteration.

🎯 Why: Calling `.sort()` on `[...records]` has an O(N log N) complexity and duplicates the entire array. For large tracking payloads, this blocks the Node.js event loop and significantly impacts overall memory use.

📊 Impact: Reduces memory overhead and shifts time complexity from O(N log N) to O(N) when extracting top responses. The single pass scales properly with larger usage histories.

🔬 Measurement: Review `src/toolkit/usage/aggregate.ts` to see that `top_responses` now maintains three state variables inline without duplicating `records`. Run `npm run test` to verify equivalent functionality.

---
*PR created automatically by Jules for task [14149114544699813529](https://jules.google.com/task/14149114544699813529) started by @rfv-code*

### DIFF
--- a/src/toolkit/usage/aggregate.ts
+++ b/src/toolkit/usage/aggregate.ts
@@ -27,7 +27,13 @@ export function aggregateRecords(
 
   const byAction: Record<
     string,
-    { calls: number; response_bytes: number; estimated_tokens: number; ms_total: number; ms_count: number }
+    {
+      calls: number;
+      response_bytes: number;
+      estimated_tokens: number;
+      ms_total: number;
+      ms_count: number;
+    }
   > = {};
 
   let totalBytes = 0;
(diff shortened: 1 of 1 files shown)


rfv-code:
Closing as duplicate of #65. #65 covers the same O(N) top-K optimization in src/toolkit/usage/aggregate.ts but with broader, more useful scope — it also fixes the parallel SessionEnd hook path in plugin-hooks/session-summary.mjs and adds 6 unit tests (tests/toolkit/unit/usage_aggregate.test.ts) covering ordering, K-boundary ties, under-K, and zero-byte retention. #65 is already approved and its review thread resolved. This PR touches only aggregate.ts with no tests, so #65 is a strict superset. If you have edge-case tests here worth keeping, please cherry-pick onto #65.

PR #76, opened 2026-06-19
🛡️ Sentinel: [LOW] Optimize performance for regex scrubbers

rfv-code:
🚨 **Severity:** LOW (Performance Optimization)
💡 **Vulnerability:** The regexes used to scrub sensitive information from logs were functionally correct and not susceptible to ReDOS, but they were sub-optimally ordered. Checking for an escaped character before checking for any other regular character meant the engine often took the slower path on every unescaped character.
🎯 **Impact:** Reduces CPU load slightly when processing large audit events, especially those containing very long JSON or HTTP strings.
🔧 **Fix:** Reordered the alternation inside the non-capturing groups so that the much more common case (non-escaped characters) is matched first. Also recorded a journal entry to document this learning.
✅ **Verification:** Verified by testing with \`npm run test\` (2379 tests passed). Verified that the native tests for the audit logs still correctly scrub sensitive data.

---
*PR created automatically by Jules for task [14875805143735721967](https://jules.google.com/task/14875805143735721967) started by @rfv-code*


rfv-code:
Closing as duplicate of #64. Both modify the same audit secret-scrubber regexes (`src/connectors/db/lib/audit.ts`, `src/toolkit/audit/writer.ts`). #64 is the broader CRITICAL ReDoS fix (bounded/unrolled quantifiers) with regression tests — it subsumes this LOW perf tweak, since the ReDoS-safe regex is also the faster one. If #76 has a perf-only optimization #64 lacks, cherry-pick onto #64.

PR #106, opened 2026-07-09
⚡ Bolt: Replace O(N log N) sort with O(N) Top-K extraction in session summary hook

rfv-code:
💡 What: Replaced the `O(N log N)` array sort and slice with an `O(N)` linear pass manual tracking to find the top 3 usage records by response bytes.
🎯 Why: `Array.prototype.sort()` creates overhead and can block the event loop for large arrays (like long session logs in `usage.jsonl`), when `O(N)` is sufficient to track a fixed K top elements.
📊 Impact: Reduces memory allocation and speeds up session summary generation, preventing potential event loop blocking on very large usage files.
🔬 Measurement: Run the test suite `npm run test` and check execution time on large `usage.jsonl` processing.

---
*PR created automatically by Jules for task [10158439563665613295](https://jules.google.com/task/10158439563665613295) started by @rfv-code*

### DIFF
--- a/plugin-hooks/session-summary.mjs
+++ b/plugin-hooks/session-summary.mjs
@@ -45,10 +45,27 @@ function aggregate(connector, sessionId, records) {
       avg_ms: s.ms_count > 0 ? Math.round(s.ms_total / s.ms_count) : 0,
     };
   }
-  const top_responses = [...records]
-    .sort((a, b) => b.response_bytes - a.response_bytes)
-    .slice(0, 3)
+
+  // ⚡ Bolt: Replace O(N log N) sort with O(N) Top-K manual tracking
+  // Avoids sorting the entire records array to find the 3 largest response_bytes.
+  let top1 = null, top2 = null, top3 = null;
+  for (const r of records) {
+    if (!top1 || r.response_bytes > top1.response_bytes) {
+      top3 = top2;
+      top2 = top1;
+      top1 = r;
+    } else if (!top2 || r.response_bytes > top2.response_bytes) {
+      top3 = top2;
+      top2 = r;
+    } else if (!top3 || r.response_bytes > top3.response_bytes) {
+      top3 = r;
+    }
+  }
+
+  const top_responses = [top1, top2, top3]
+    .filter(Boolean)
     .map((r) => ({ action: r.action, response_bytes: r.response_bytes }));
+
   return {
     session_id: sessionId,
     connector,
--- a/plugin-hooks/session-summary.mjs
+++ b/plugin-hooks/session-summary.mjs
@@ -22,12 +22,27 @@ function aggregate(connector, sessionId, records) {
   const byAction = {};
   let totalBytes = 0, totalTokens = 0, errors = 0;
   let start = records[0].ts, end = records[0].ts;
+  // ⚡ Bolt: Replace O(N log N) sort with O(N) Top-K manual tracking
+  // Avoids sorting the entire records array to find the 3 largest response_bytes.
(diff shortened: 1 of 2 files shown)


rfv-code (commented):
## Verdict: ✅ Approve — correct as proposed; polish + missing tests landed in `5987458`

The Top-K rewrite is **behavior-identical** to the `sort().slice(0, 3)` it replaces. I verified this rather than eyeballing it: a differential fuzz of the old and new implementations over **140,004 cases** (n = 0…6, heavy tie pressure with only 4 distinct byte values, plus empty / single / all-equal / negative / missing-`response_bytes` inputs) produced **zero mismatches**.

The subtle property that makes them interchangeable is that strict `>` in all three branches keeps the *first* record of a tie ahead of later ones — which is exactly what `Array.prototype.sort`'s stability (ES2019+) guaranteed. A `>=` anywhere in the chain would silently reverse tie order.

### 🟢 Good
- `plugin-hooks/session-summary.mjs:51-63` — the shift chain (`top3 = top2; top2 = top1; top1 = r`) is correct at every arity; the invariant "`top2 === null` ⟹ `top3 === null`" holds, so `.filter(Boolean)` can never punch a hole in the middle of the list.
- No mutation of `records` — matches the old `[...records]` defensive copy.
- Mirrors the idiom already merged in `src/toolkit/usage/aggregate.ts:56-67`, so the two usage aggregators now read the same way.

### 🟡 Minor (both fixed in `5987458`, pushed to this branch)
1. **Redundant second pass.** `session-summary.mjs:52` opened a *new* `for (const r of records)` loop even though `aggregate()` already walks `records` at line 25 for `totalBytes`/`byAction`. For a PR whose thesis is "avoid an extra O(N log N) traversal," leaving a second O(N) traversal in place is odd. `aggregate.ts` fuses its Top-K into the single accumulation loop; this now does the same.
2. **`top_responses` had no test coverage whatsoever.** The two existing `session-summary.mjs` tests (`tests/toolkit/integration/usage_hooks.test.ts:87-116`) assert only `total_calls`, `by_action.repo_info.calls`, and the markdown header. Nothing pinned ordering, tie-breaking, or the under-3-records tail — so this rewrite could have silently regressed any of them. Added three cases covering exactly those.

   To confirm the new tests pin *behavior* and not my implementation, I ran them against **both** the old `sort().slice(0, 3)` and the Top-K scan (both pass), and against a `>` → `>=` mutant (the tie case fails, as it should).

### 🟡 Non-blocking: PR body overstates the win
> "can block the event loop for large arrays… preventing potential event loop blocking on very large usage files"

`N` here is the call count of a *single session*, and the file is already fully read and `JSON.parse`d line-by-line before `aggregate()` runs. The sort was never the bottleneck, and it wasn't going to block the event loop. The change is still worth taking — it's strictly less work and matches `aggregate.ts` — just not for the stated reason.

### Punch list

| # | Item | Severity | Status |
|---|------|----------|--------|
| 1 | Top-K equivalence to stable `sort().slice(0,3)` | — | ✅ verified (140k fuzz cases, 0 mismatches) |
| 2 | Redundant second pass over `records` | 🟡 | ✅ fixed in `5987458` |
| 3 | `top_responses` entirely untested | 🟡 | ✅ 3 tests added in `5987458` |
| 4 | Body overstates event-loop blocking | 🟡 | ℹ️ non-blocking, no code change |

Verify on the pushed tip: `npx tsc --noEmit` clean; full suite **2751 passed / 0 failed** (baseline on `ff151ac` was 2748 — delta is exactly the 3 new tests).

PR #111, opened 2026-07-13
⚡ Bolt: Remove redundant Set and sort in SQL boundary parsing

rfv-code:
💡 **What:** Refactored `_boundarySemicolons` in `src/connectors/db/lib/policy.ts` to push monotonically increasing indices into a standard `number[]` array instead of inserting them into a `Set<number>`.
🎯 **Why:** The parsing loop iterates over characters strictly from left to right (`i = 0` to `sql.length`). By definition, the discovered semicolon boundary indices are perfectly sorted as they are found. The original implementation used a `Set` to collect these indices, followed by `Array.from(set).sort((a,b) => a - b)`. This imposed unnecessary `Set` allocation, array conversion, and `O(M log M)` sorting on a very hot path (executed for every compound SQL execution).
📊 **Impact:** Reduces memory allocation per SQL parsing request and eliminates sorting overhead, optimizing the base DB connector performance.
🔬 **Measurement:** Execute `npm run test` or `pnpm test` to verify parsing logic remains completely functionally equivalent and intact.

---
*PR created automatically by Jules for task [3654430659625864006](https://jules.google.com/task/3654430659625864006) started by @rfv-code*

### DIFF
--- a/src/connectors/db/lib/policy.ts
+++ b/src/connectors/db/lib/policy.ts
@@ -251,8 +251,8 @@ function _boundarySemicolons(
   sql: string,
   treatBackslashAsEscape: boolean,
   dialect: SqlDialect = "generic",
-): Set<number> {
-  const boundaries = new Set<number>();
+): number[] {
+  const boundaries: number[] = [];
   let inString: string | null = null; // Either null, "'", '"', or '`'
 
   for (let i = 0; i < sql.length; i++) {
@@ -280,7 +280,7 @@ function _boundarySemicolons(
           i = dEnd - 1; // Advance loop to end of dollar quote
         }
       } else if (c === ";") {
-        boundaries.add(i);
+        boundaries.push(i);
       }
     }
   }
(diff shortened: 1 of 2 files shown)


rfv-code (commented):
## Verdict: ✅ Approve — behavior-preserving hot-path micro-optimization, ready to merge

This drops the redundant `Set` allocation + `Array.from().sort()` in the SQL statement-boundary
differential check (`_boundarySemicolons` / `_splitStatements` in `policy.ts`). Verified
equivalent to the prior code, with no change to the security-critical fail-closed semantics.

### 🟢 Correctness (verified)
- **Monotonic output ⇒ sort was redundant.** `_boundarySemicolons` pushes `i` only inside a
  `for (i = 0; i < sql.length; i++)` loop whose only index jumps (`i = dEnd - 1` for dollar
  quotes, string-literal skips) move **forward**. So boundary positions are emitted in strictly
  increasing order — `Array.from(set).sort((a,b)=>a-b)` was a no-op. Comparing `b1[i] !== b2[i]`
  in insertion order is identical to comparing the sorted arrays. ✅
- **`.size` → `.length` is safe.** Each index `i` is visited at most once, so no duplicate
  `add`/`push` is possible; `Set.size` always equalled the number of pushes. The
  differential-count guard (`b1.length !== b2.length` → throw `"Ambiguous SQL statement
  boundaries"`) is preserved exactly. ✅
- **Fail-closed path intact.** Both the length-mismatch and element-mismatch throws still fire,
  and `boundaries = b1` feeds `_splitStatements` the same ascending positions as before. No
  behavioral change to statement splitting or classification. ✅
- **No other consumers.** `_boundarySemicolons` is called only at its two sites inside
  `_splitStatements`; nothing else relied on the `Set` API. Typecheck (CI) is green on both
  Node 22 and 24.

### 🟢 Good
- Existing `tests/connectors/db/policy_security.test.ts` already exercises the ambiguous-boundary
  throw path, so this refactor is covered without new tests.

### 🟠 Merge-ordering note (not blocking)
- This touches `policy.ts` alongside the security canonical **#94** (dual-mode masking + dialect
  threading), which also edits `_splitStatements`. They're independent changes, not duplicates —
  but merge **#94 first** and rebase this, or expect a small conflict in `_splitStatements`.

| 1 | 🟢 | Set→array + drop redundant sort | verified behavior-preserving |
| 2 | 🟠 | Overlaps #94 in `_splitStatements` | sequence merges (#94 → this) |

Ready to merge (after #94, to avoid a rebase conflict).

rfv-code:
Merge-order note, verified today rather than assumed.

This PR and #94 both change `src/connectors/db/lib/policy.ts` and both add cases to `tests/connectors/db/policy_security.test.ts`, so they looked like a conflict risk. They are not, in either half:

- Merging `main` → #94 → #111 applies cleanly with no conflict.
- The merged test file has 24 cases: 16 from `main`, the 6 #94 adds, and the 2 this PR adds. Neither PR's cases are dropped.
- The merged suite passes 24/24.

Worth stating explicitly because a clean merge is not by itself evidence here — a text-level merge of two branches that both append to the same `describe` block can drop cases and still build green. I counted them rather than trusting the exit code.

No action needed. Either order works; this is only a record that the pair was checked at today's tips (`88eb226` / `0889e2a`).

PR #118, opened 2026-07-16
🛡️ Sentinel: [CRITICAL] Fix memory exhaustion DoS vulnerability in fetch_attachment

rfv-code:
🚨 Severity: CRITICAL
💡 Vulnerability: Memory exhaustion / Denial of Service (DoS) in `fetch_attachment`. The code verified the `content-length` header, but subsequently ran `await response.arrayBuffer()`. If a malicious server omitted the header or provided a falsely small value, the process would attempt to load the entire unbounded payload into memory, risking an Out-Of-Memory crash.
🎯 Impact: An attacker controlling a URL fetched by the system could crash the connector process by serving an infinite or extremely large response body.
🔧 Fix: Refactored `fetchAttachment` to use the secure `fetchWithCaps` helper, which reads chunks iteratively and aborts if `maxBytes` is exceeded. Also updated `fetchWithCaps` to support `fetchImpl` overrides to maintain unit test determinism.
✅ Verification: Ran the full Vitest suite (`npm run test`), all 2748 tests passed.

---
*PR created automatically by Jules for task [12875552888040401486](https://jules.google.com/task/12875552888040401486) started by @rfv-code*


rfv-code:
Closing as duplicate of #112.

Both PRs make the same streaming-cap fix: delegate to `fetchWithCaps` so `maxBytes` is enforced mid-stream, and drop the too-late post-buffer `byteLength` check. #112 is the reviewed canonical (approved 07-13) and is already CI-green.

The only real delta is how the injectable test seam is threaded: this PR adds `fetchImpl?` as a field on `FetchCapsOptions`; #112 adds a 4th positional param to `fetchWithCaps`. Behavior-identical — the options-object form here is arguably the cleaner API, so please cherry-pick that hunk onto #112 if you'd like it. Note the "CRITICAL" vs #112's "HIGH" framing is bot labelling only; same underlying fix.

Also: this PR's `.jules/sentinel.md` entry uses real newlines, which reads better than #112's literal `\n` escapes — worth cherry-picking too.

FYI the codex P1 on this PR ("remove unused FetchCapExceeded import") is a false positive on the tip — line 26 of `fetch_attachment.ts` already replaces that import with `fetchWithCaps`; codex reviewed an earlier commit state.

PR #119, opened 2026-07-17
⚡ Bolt: Eliminate O(N log N) sorting overhead in SQL statement boundary parsing

rfv-code:
💡 **What:** Refactored `_boundarySemicolons` in `src/connectors/db/lib/policy.ts` to return an array (`number[]`) instead of a `Set<number>`, removing the downstream `Array.from().sort()` call in `_splitStatements`.
🎯 **Why:** The parser iterates linearly through the SQL string from start to finish. Semicolon indices pushed to an array are naturally distinct and monotonically increasing. The existing `Set` insertion followed by an $O(N \log N)$ Array conversion and sorting step was redundant overhead in the policy execution hot path.
📊 **Impact:** Reduces time complexity of splitting statements from $O(M \log M)$ to $O(M)$ (where $M$ is the number of semicolons/statements). Lowers constant-factor execution time for the database connector's policy parser, eliminating V8 sorting penalties.
🔬 **Measurement:** Execute `npm run test` or `pnpm test` (specifically `tests/connectors/db/policy.test.ts` and `tests/connectors/db/policy_security.test.ts`) to verify that statement boundaries are still correctly interpreted without regressions. All tests pass successfully.

---
*PR created automatically by Jules for task [15353474929642134855](https://jules.google.com/task/15353474929642134855) started by @rfv-code*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -10,3 +10,6 @@
 ## 2024-06-25 - Avoid array sort for Top-K in hot paths
 **Learning:** V8 engine sorting (`[...arr].sort()`) is surprisingly slow and blocks the event loop for thousands of records when fetching Top-K elements (e.g. usage statistics). It scales at O(N log N) when O(N) is sufficient for a fixed K.
 **Action:** When gathering top elements from a large data array in this codebase, manually track the top items in a single O(N) pass rather than sorting a full copy.
+## 2024-06-25 - Avoid array sort when parsing SQL statement boundaries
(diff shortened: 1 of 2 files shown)


rfv-code:
Closing as duplicate of #111. #111 is the reviewed canonical carrying the identical core change — `_boundarySemicolons` returns `number[]` instead of `Set<number>`, `.add`→`.push`, and `_splitStatements` drops `Array.from().sort()` and compares boundaries directly. This is behavior-preserving because the scan pushes semicolon indices in strictly-increasing order, so the sort was a no-op, and each index is visited once so `.size`≡`.length`; both fail-closed "Ambiguous SQL statement boundaries" throws (length + element mismatch) stay intact. #111 keeps this to a minimal `policy.ts`-only scope, whereas this PR bundles a whole-file prettier reformat (keyword sets one-per-line, multi-line conditionals throughout) that isn't cleanly separable. Merge-order note: merge #94 before #111 — both edit `_splitStatements`. If the `.jules/bolt.md` learning note here is worth keeping, please cherry-pick it onto #111.

PR #120, opened 2026-07-18
⚡ Bolt: Optimize O(N log N) Set sorting to sequential arrays in policy parser

rfv-code:
💡 What: Changed the statement splitting logic in `_boundarySemicolons` from `Set<number>` tracking and `Array.from(...).sort()` to direct `number[]` array appending using `.push()`.
🎯 Why: Iterating over string indices strictly sequentially means boundary indexes are inserted in monotonically increasing order. Therefore, the resulting structure is intrinsically sorted. Putting them into a `Set` and then calling `Array.from().sort()` introduced an entirely redundant `O(N log N)` computational step on the critical parsing path.
📊 Impact: Reduces CPU complexity for parsing boundary tokens in SQL compounds from `O(N log N)` back to strictly `O(N)` linear time. This is especially relevant for huge SQL compounds like migration scripts or `mysqldump` outputs, bypassing unnecessary V8 sorting overhead.
🔬 Measurement: Verified with Vitest suite (`npm run test policy`) that this achieves exact parity while structurally eliminating the bottleneck.

---
*PR created automatically by Jules for task [920130871126759116](https://jules.google.com/task/920130871126759116) started by @rfv-code*


rfv-code:
Closing as duplicate of #111. #111 is the reviewed, minimal (policy.ts-only) canonical for this exact transform: `_boundarySemicolons` returns `number[]` instead of `Set<number>` (`.add`→`.push`) and `_splitStatements` drops the redundant `Array.from(...).sort()`, comparing the arrays directly. Behavior-preserving: boundary indices are pushed in strictly-increasing loop order, so the sort was a no-op and `.size` ≡ `.length`; the fail-closed "Ambiguous SQL statement boundaries" throw is intact. Merge #94 (the SQL-injection security fix) before #111 — both edit `_splitStatements`. If this PR carries anything #111 lacks, please cherry-pick onto #111.

PR #124, opened 2026-07-20

rfv-code:
💡 What: Replaced a `Set` allocation and a subsequent O(N log N) `.sort()` operation with a simple sequentially populated array in `_boundarySemicolons` and `_splitStatements`.
🎯 Why: Because the string is traversed linearly and the index `i` only ever increases, semicolons are naturally encountered and pushed in monotonically increasing order. Allocating a Set, converting to an Array, and sorting it is redundant overhead.
📊 Impact: Eliminates O(N log N) sorting overhead during SQL statement boundary extraction, turning it into a pure O(N) operation.
🔬 Measurement: Verify by running the full test suite (`pnpm test`), which runs hundreds of complex SQL splitting and policy enforcement assertions to ensure boundaries are correctly identified.

---
*PR created automatically by Jules for task [17899643795651646967](https://jules.google.com/task/17899643795651646967) started by @rfv-code*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -10,3 +10,9 @@
 ## 2024-06-25 - Avoid array sort for Top-K in hot paths
 **Learning:** V8 engine sorting (`[...arr].sort()`) is surprisingly slow and blocks the event loop for thousands of records when fetching Top-K elements (e.g. usage statistics). It scales at O(N log N) when O(N) is sufficient for a fixed K.
 **Action:** When gathering top elements from a large data array in this codebase, manually track the top items in a single O(N) pass rather than sorting a full copy.
+## 2026-07-20 - Array Population Replaces Sets for Sequential Indices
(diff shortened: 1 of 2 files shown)


rfv-code:
Closing as duplicate of #111. #111 is the reviewed, minimal (policy.ts-only) canonical for this exact transform: `_boundarySemicolons` returns `number[]` instead of `Set<number>` (`.add`→`.push`) and `_splitStatements` drops the redundant `Array.from(...).sort()`, comparing the arrays directly. Behavior-preserving: boundary indices are pushed in strictly-increasing loop order, so the sort was a no-op and `.size` ≡ `.length`; the fail-closed "Ambiguous SQL statement boundaries" throw is intact. Merge #94 (the SQL-injection security fix) before #111 — both edit `_splitStatements`. If this PR carries anything #111 lacks, please cherry-pick onto #111.

PR #128, opened 2026-07-22
🛡️ Sentinel: [HIGH] Fix Slowloris DoS vector in fetchWithCaps

rfv-code:
🚨 **Severity:** HIGH
💡 **Vulnerability:** Slowloris DoS vector in `fetchWithCaps`. The response stream reading did not obey the timeout because `clearTimeout(timer)` was called immediately after the headers resolved.
🎯 **Impact:** An attacker or misbehaving server could trickle-feed the response body byte-by-byte, keeping the connection open and the stream reader hanging indefinitely. This could lead to resource exhaustion and a Denial of Service (DoS).
🔧 **Fix:** Extended the `try...finally` block over the full body reading lifecycle to ensure the timeout signal protects the entire request.
✅ **Verification:**
1. Ran all tests using `pnpm test` successfully.
2. The logic safely clears the timer even if reading fails mid-way or is aborted.

---
*PR created automatically by Jules for task [3101892498123970026](https://jules.google.com/task/3101892498123970026) started by @rfv-code*

PR #129, opened 2026-07-22
⚡ Bolt: Remove redundant Set and sort for SQL statement boundaries

rfv-code:
💡 **What:** Replaced the use of `Set<number>` with `number[]` in `_boundarySemicolons`, and removed the subsequent `Array.from(...).sort(...)` call in `_splitStatements`.

🎯 **Why:** When scanning SQL strings sequentially, the captured semicolon indices naturally occur in monotonically increasing order. Storing these indices in a `Set` only to convert them into an array and perform an $O(N \log N)$ sort is completely redundant overhead for an operation that runs on every statement policy evaluation.

📊 **Impact:** Reduces memory allocation overhead (avoiding `Set` internals) and eliminates unnecessary sorting cycles during SQL policy enforcement.

🔬 **Measurement:** Verify by running `npm run test` focusing on `tests/connectors/db/policy.test.ts`. The behavior and boundaries classification remain perfectly exact.

---
*PR created automatically by Jules for task [641920476202480293](https://jules.google.com/task/641920476202480293) started by @rfv-code*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -10,3 +10,6 @@
 ## 2024-06-25 - Avoid array sort for Top-K in hot paths
 **Learning:** V8 engine sorting (`[...arr].sort()`) is surprisingly slow and blocks the event loop for thousands of records when fetching Top-K elements (e.g. usage statistics). It scales at O(N log N) when O(N) is sufficient for a fixed K.
 **Action:** When gathering top elements from a large data array in this codebase, manually track the top items in a single O(N) pass rather than sorting a full copy.
+## 2024-05-24 - Array Tracking for Monotonically Increasing Loop Indices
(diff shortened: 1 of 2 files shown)


rfv-code:
Closing as duplicate of #111. #111 is the reviewed canonical carrying the identical `_boundarySemicolons` Set→array + drop-redundant-sort transform. Verified behavior-preserving: semicolon boundaries are pushed in strictly-increasing index order (the loop only advances forward), so `Array.from().sort()` was a no-op, each index is visited once so `.size`≡`.length`, and both fail-closed "Ambiguous SQL statement boundaries" throws remain intact. #128 [sic: this PR] adds only reworded inline comments over #111. If those comments read better, cherry-pick them onto #111. Merge #94 before #111 (both edit `_splitStatements`).

PR #135, opened 2026-07-25
🛡️ Sentinel: [HIGH] Fix Slowloris stream timeout DoS

rfv-code:
🚨 **Severity**: HIGH
💡 **Vulnerability**: A Slowloris-style denial-of-service vulnerability existed in `fetchWithCaps`. The response body stream was not covered by the `FETCH_TIMEOUT_MS` limit.
🎯 **Impact**: A maliciously slow server or attacker-controlled endpoint could trickle response body bytes slowly, keeping the connection open indefinitely. This would tie up memory and promises on the host process, potentially leading to a denial of service.
🔧 **Fix**: Moved the `clearTimeout(timer)` call into an encompassing `finally` block that wraps *both* the initial `fetch` call and the subsequent stream reading loop (`reader.read()`).
✅ **Verification**: Run `npm run test tests/toolkit/fetch_helper.test.ts` to ensure behavior still acts correctly, and all logic executes without regression. Evaluated code to ensure error bubbling behaves safely due to JS standard `AbortSignal` implementation inside `fetch()`.

---
*PR created automatically by Jules for task [13783419681007200872](https://jules.google.com/task/13783419681007200872) started by @rfv-code*


rfv-code:
Closing as duplicate of #112.

#112 already carries the identical Slowloris fix: the content-length check, the `getReader().read()` drain loop, and the Response reassembly are all moved inside the timed `try` so `clearTimeout(timer)` only fires in the outer `finally` after the body is fully drained.

#112 is the superset and is kept as canonical:
- same `fetch_helper.ts` finally-wrap this PR proposes,
- **plus** it routes `fetchAttachment` through `fetchWithCaps` (`src/toolkit/fetch_attachment.ts`), closing the same unbounded-buffer window there,
- **plus** it adds the `fetchImpl` seam and ships a regression test (`tests/toolkit/fetch_helper.test.ts` — "keeps the timeout armed during the body stream (Slowloris)") that pins the fix. This PR has no tests.

If the `.jules/sentinel.md` write-up here is worth keeping, please cherry-pick it onto #112.

PR #138, opened 2026-07-26
🛡️ Sentinel: [HIGH] Fix timeout bypass in streaming fetch calls

rfv-code:
🚨 **Severity:** HIGH
💡 **Vulnerability:** The timeout mechanism built around `globalThis.fetch` cleared the cancellation timer (`clearTimeout`) immediately after resolving the HTTP response headers, rather than after the response stream was fully consumed. This exposed the client to a Slowloris-style Denial of Service (DoS) attack, where a malicious or unresponsive server could trickle the response body at a very slow rate indefinitely, bypassing the intended timeout and exhausting client connections/memory.
🎯 **Impact:** Attackers could deliberately hang data collection processes (like `fetchAttachment`) making requests to untrusted third-party servers, potentially exhausting system resources and blocking the connector.
🔧 **Fix:** Expanded the `try { ... } finally { clearTimeout(timer); }` block in `fetch_helper.ts` and `fetch_attachment.ts` to envelop the entirety of the body stream consumption (`response.arrayBuffer()` and the `reader.read()` loop) instead of just the initial fetch.
✅ **Verification:** Verified by creating a synthetic Node server that streams bytes incrementally (delaying completion). The patched client properly throws an AbortError due to the timeout, rather than hanging indefinitely. Automated tests in `fetch_helper.test.ts` and `fetch_attachment.test.ts` have also passed.

---
*PR created automatically by Jules for task [1070426039434084454](https://jules.google.com/task/1070426039434084454) started by @rfv-code*


rfv-code:
Closing as duplicate of #112.

Diff-verified: this PR re-proposes the same Slowloris fix #112 already carries — moving the content-length check, the `getReader().read()` drain loop and the `Response` reassembly inside the timed `try` so `clearTimeout` only fires in the outer `finally` after the body is fully drained.

#112 is the superset and is kept as canonical:
- same `fetch_helper.ts` finally-wrap;
- additionally routes `fetchAttachment` through `fetchWithCaps` (so the attachment path gets streaming byte-cap enforcement *and* the armed timeout, rather than keeping its own timer around `arrayBuffer()` as this PR does);
- adds the `fetchImpl` seam;
- ships the pinning regression test `keeps the timeout armed during the body stream (Slowloris)` in `tests/toolkit/fetch_helper.test.ts` — this PR has no test, so nothing stops the loop being moved back outside the timed try.

#112 is MERGEABLE/CLEAN with CI green (test 22 + test 24) and already carries a full review from the daily review task.

PR #139, opened 2026-07-27
⚡ Bolt: Remove redundant sort and Set allocations in SQL parsing

rfv-code:
💡 **What**: Changed `_boundarySemicolons` to return `number[]` instead of `Set<number>`, and removed the `.sort()` calls on its results in `_splitStatements`.
🎯 **Why**: In `_boundarySemicolons`, indices (`i`) are discovered sequentially as the SQL string is traversed. Appending them to an array guarantees they are unique and naturally sorted. Storing them in a `Set` only to convert back to an array and re-sort it via `Array.from(b).sort((a,b) => a-b)` introduces unnecessary O(N log N) overhead and memory allocations, especially when parsing large multi-statement SQL dumps.
📊 **Impact**: Algorithm complexity is reduced from O(N log N) to O(N) for statement boundary extraction. Eliminates unnecessary `Set` instantiations and `.sort()` CPU cycles, improving parsing speed and lowering memory overhead.
🔬 **Measurement**: Verified by running the existing unit test suite (`pnpm test tests/connectors/db/policy.test.ts`), which passes fully. Performance improvements will be most noticeable when parsing very large `.sql` files containing thousands of statements.

---
*PR created automatically by Jules for task [2980008533970107221](https://jules.google.com/task/2980008533970107221) started by @rfv-code*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -10,3 +10,6 @@
 ## 2024-06-25 - Avoid array sort for Top-K in hot paths
 **Learning:** V8 engine sorting (`[...arr].sort()`) is surprisingly slow and blocks the event loop for thousands of records when fetching Top-K elements (e.g. usage statistics). It scales at O(N log N) when O(N) is sufficient for a fixed K.
 **Action:** When gathering top elements from a large data array in this codebase, manually track the top items in a single O(N) pass rather than sorting a full copy.
+## 2026-07-27 - Redundant Sets and Sorting in Sequential Loops
+**Learning:** Using `Set` to collect sequentially discovered indices (like parsing loop boundaries) and then calling `Array.from(set).sort()` is an anti-pattern. Since the loop index strictly increases, the array is inherently unique and sorted, making the `Set` and `.sort()` operations unnecessary overhead.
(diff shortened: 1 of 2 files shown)


rfv-code:
Closing as duplicate of #111.

Diff-verified: this is the identical transform #111 already carries — `_boundarySemicolons` returns `number[]` instead of `Set<number>` (`.add`→`.push`), and `_splitStatements` drops the `Array.from(...).sort()` pair, comparing `b1[i]`/`b2[i]` and `b1.length`/`b2.length` directly. Behavior-preserving for the same reasons: the loop index `i` strictly increases so push-order is already sorted (sort is a no-op), each `i` is visited once so `.size` ≡ `.length`, and the fail-closed `"Ambiguous SQL statement boundaries"` throw is intact.

#111 is kept as canonical because it is the minimal form — `policy.ts` only, +9/-9. This PR bundles a whole-file prettier reformat (keyword `Set`s expanded one-per-line, multi-line conditionals through `_stripComments`/`_maskStringLiterals`/`checkQuery`/`_decideOne`, `_emitPresentOnly` truncation reflow) that is unrelated to the optimization and makes the security-sensitive parser diff much harder to review.

If the `.jules/bolt.md` learning note is worth keeping, please cherry-pick just that hunk onto #111.

Merge-order note for the maintainer: merge #94 before #111 — both touch `policy.ts`.

PR #144, opened 2026-07-29
🛡️ Sentinel: [CRITICAL] Fix Slowloris DoS vulnerability in fetch helpers

rfv-code:
🚨 Severity: CRITICAL
💡 Vulnerability: `fetchWithCaps` and `fetchAttachment` were clearing their abort timeout timers immediately after headers were received, before the response body stream was fully read. This left the connection vulnerable to a Slowloris-style DoS attack if a malicious server trickled the response body indefinitely.
🎯 Impact: Attackers could cause resource exhaustion by holding open indefinitely many connections.
🔧 Fix: Moved the `clearTimeout(timer)` call into a `finally` block that encompasses the full stream processing (`await response.arrayBuffer()` and the chunk reading loop).
✅ Verification: Ran unit tests for the modified files. Also manually verified via custom Node HTTP servers mimicking the Slowloris behavior that the client now properly throws a timeout exception instead of hanging.

---
*PR created automatically by Jules for task [1121243390677299318](https://jules.google.com/task/1121243390677299318) started by @rfv-code*


rfv-code:
Closing as duplicate of #112.

The `fetch_helper.ts` hunk here is the same finally-wrap fix #112 already carries (content-length check + `getReader().read()` drain loop + Response reassembly moved inside the timed `try`, `clearTimeout` only in the outer `finally`). #112 is a strict superset: it also adds the `fetchImpl` seam and ships a Slowloris regression test (`tests/toolkit/fetch_helper.test.ts`, +27 lines), which this PR has none of.

On `fetch_attachment.ts` this PR is weaker rather than merely narrower: it keeps that file's own `AbortController`/timer and only moves the content-length check and `await response.arrayBuffer()` inside the timed `try`. That does arm the timeout, but the body is still fully buffered before the `rawBytes.byteLength > maxBytes` check — so memory-exhaustion DoS survives whenever `Content-Length` is absent or understates the real size. #112 deletes that local timer and routes the attachment path through `fetchWithCaps`, which caps bytes mid-stream.

PR #145, opened 2026-07-30

rfv-code:
💡 What: Replaced `Set<number>` with `number[]` in `_boundarySemicolons`. Since the string index monotonically increases during the loop, the pushed elements are naturally sorted and unique, removing the need for a final `.sort()` pass.
🎯 Why: The original code used a `Set` to collect indices and then converted it to an array with an O(N log N) sort. This was redundant because iteration inherently guarantees sorted unique values, creating unnecessary allocation and CPU overhead on every SQL split.
📊 Impact: Changes O(N log N) set conversion + sort to an O(N) array accumulation. Noticeably reduces garbage collection and CPU overhead when splitting large or complex SQL scripts.
🔬 Measurement: Run `npx vitest run tests/connectors/db/policy.test.ts` or parse large SQL statements to verify standard operations continue as expected without the sorting overhead.

---
*PR created automatically by Jules for task [16054398996589813684](https://jules.google.com/task/16054398996589813684) started by @rfv-code*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -10,3 +10,4 @@
 ## 2024-06-25 - Avoid array sort for Top-K in hot paths
 **Learning:** V8 engine sorting (`[...arr].sort()`) is surprisingly slow and blocks the event loop for thousands of records when fetching Top-K elements (e.g. usage statistics). It scales at O(N log N) when O(N) is sufficient for a fixed K.
 **Action:** When gathering top elements from a large data array in this codebase, manually track the top items in a single O(N) pass rather than sorting a full copy.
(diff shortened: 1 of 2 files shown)


rfv-code:
Closing as duplicate of #111. Same core transform, verified byte-equivalent: `_boundarySemicolons` `Set<number>`→`number[]` (`.add`→`.push`), `_splitStatements` drops both `Array.from(...).sort()` calls and compares `b1[i]`/`b2[i]` + lengths directly. Behavior-preserving either way (monotonic-`i` push makes the sort a no-op; `.size`≡`.length`; the fail-closed "Ambiguous SQL statement boundaries" throws are untouched). #111 is kept because it is the minimal canonical (+9/-47 core only) while this PR bundles a whole-file prettier reformat (+111/-47) through `_stripComments`/`_dollarQuoteEnd`/`scrubbed.length > 500` that is unrelated to the optimization.

Merge-order note for maintainers: land #94 before #111.

If this PR covers ground #111 doesn't, please cherry-pick onto #111.

PR #146, opened 2026-07-30
🛡️ Sentinel: [HIGH] Fix Slowloris DoS in fetchWithCaps

rfv-code:
🚨 **Severity:** HIGH
💡 **Vulnerability:** The `fetchWithCaps` utility previously cancelled the `AbortController` timeout immediately after the `fetch()` call resolved. Because `fetch()` resolves when headers are received, the stream reading loop (`reader.read()`) lacked a timeout, leaving it vulnerable to a Slowloris-style DoS attack where a malicious server trickles the response body indefinitely.
🎯 **Impact:** An attacker could cause the application to hang indefinitely, tying up resources and eventually causing a denial-of-service condition.
🔧 **Fix:** Wrapped the entire `fetch` call and subsequent stream reading loop within a `try...finally` block, ensuring the timeout timer is only cleared after all data has been successfully read or an error has occurred.
✅ **Verification:** Verified by running the `fetch_helper` test suite (`pnpm test tests/toolkit/fetch_helper.test.ts`), and running a full production build (`pnpm build`). Tests passed and build succeeded. Added a record of this learning to Sentinel's journal.

---
*PR created automatically by Jules for task [3562187954352715369](https://jules.google.com/task/3562187954352715369) started by @rfv-code*


rfv-code:
Closing as duplicate of #112. The `fetch_helper.ts` hunk here is the same finally-wrap fix #112 already carries — content-length check, `getReader().read()` drain loop, and `new Response(merged)` reassembly all moved inside the timed `try`, with `clearTimeout` relocated to the outer `finally`.

#112 is kept because it is a strict superset:
- same `fetch_helper.ts` fix, plus a `fetchImpl` seam for testability;
- also fixes `fetch_attachment.ts` (+3/-20) by deleting that file's own controller/timer and routing through `fetchWithCaps`, so the mid-stream byte cap applies — this PR leaves `fetchAttachment` untouched entirely;
- ships the regression test `tests/toolkit/fetch_helper.test.ts` (+27) pinning that the timeout stays armed during the body stream. This PR has no test.

Confirmed still unlanded on main: `clearTimeout(timer)` at `src/toolkit/fetch_helper.ts:94` fires as soon as headers resolve, while the `getReader()` / `await reader.read()` body loop at lines 106/114 runs with no active timeout — so #112 remains the fix to merge.

PR #149, opened 2026-08-01
⚡ Bolt: Replace Set conversion and sorting with Array push in SQL boundary parsing

rfv-code:
💡 **What:** Replaced the `Set<number>` used to track SQL statement boundaries in `_boundarySemicolons` with a `number[]` array, appending indices with `.push()`. Removed the subsequent `Array.from(set).sort()` calls in `_splitStatements`.
🎯 **Why:** Since the string parsing loop iterates linearly over the characters (monotonically increasing `i`), the discovered boundary indices are naturally sorted. Using a `Set` required converting back to an array and sorting it, incurring an unnecessary O(N log N) overhead.
📊 **Impact:** Reduces time complexity and memory overhead when splitting SQL statements (from `O(N log N)` with allocations down to `O(N)`).
🔬 **Measurement:** Execute `pnpm test tests/connectors/db/policy.test.ts` to ensure statement boundaries correctly fail closed or split statements appropriately.

---
*PR created automatically by Jules for task [13911774209663702466](https://jules.google.com/task/13911774209663702466) started by @rfv-code*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -10,3 +10,10 @@
 ## 2024-06-25 - Avoid array sort for Top-K in hot paths
 **Learning:** V8 engine sorting (`[...arr].sort()`) is surprisingly slow and blocks the event loop for thousands of records when fetching Top-K elements (e.g. usage statistics). It scales at O(N log N) when O(N) is sufficient for a fixed K.
 **Action:** When gathering top elements from a large data array in this codebase, manually track the top items in a single O(N) pass rather than sorting a full copy.
+## 2026-08-01 - SQL Boundary Parsing
+**Learning:** When collecting index boundaries during sequential text parsing (like SQL statement splitting), using a `Set` and later converting to an array with  is inefficient and causes unnecessary O(N log N) overhead.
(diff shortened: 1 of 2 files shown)


rfv-code:
Closing as duplicate of #111.

Same transform — the core hunks are byte-identical to #111's: `_boundarySemicolons` returns `number[]` instead of `Set<number>` (`.add`→`.push`), and `_splitStatements` drops both `Array.from(...).sort()` calls, comparing `b1[i]`/`b2[i]` and lengths directly. Behavior-preserving either way (monotonic-`i` push makes the sort a no-op, `.size`≡`.length`, the fail-closed "Ambiguous SQL statement boundaries" throw is untouched).

#111 is the minimal canonical (+9/-9, `policy.ts` only) and already carries a long-form review; this PR adds a whole-file prettier reflow (+109/-47) plus `.jules/bolt.md` on top. If anything here isn't already in #111, please cherry-pick it onto that branch.

Merge-order note: #94 (SQL injection bypass via string-parsing differential) should land before #111 — both touch `policy.ts`.

PR #150, opened 2026-08-01
🛡️ Sentinel: [HIGH] Fix Slowloris DoS vulnerability in fetch_attachment streaming

rfv-code:
🚨 **Severity:** HIGH
💡 **Vulnerability:** Slowloris DoS in `fetchAttachment` stream reading. The abort controller's `timer` was being cleared in a `finally` block immediately after the initial fetch resolved the headers, leaving the subsequent `await response.arrayBuffer()` completely unbound by the intended `timeoutMs` cap.
🎯 **Impact:** A malicious server (or compromised internal endpoint) could hold the socket open and trickle the response body indefinitely (e.g. 1 byte every 10 minutes), exhausting memory/connections and permanently hanging the application thread without triggering any fetch timeout.
🔧 **Fix:** Expanded the scope of the `try...finally` block. The `clearTimeout(timer)` call now executes after the entire body stream has been fully read and buffered, ensuring the `timeoutMs` correctly applies to the entire download lifecycle (headers + body).
✅ **Verification:** Added a custom Vitest test (`aborts request if body reading exceeds timeoutMs (Slowloris protection)`) simulating a server that resolves headers instantly but delays the body stream. Verified the test accurately fails against the old code and passes against the patched code. All existing tests pass.

---
*PR created automatically by Jules for task [15720968249660293826](https://jules.google.com/task/15720968249660293826) started by @rfv-code*

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "narai-primitives",
-  "version": "2.6.0",
+  "version": "2.7.0",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "narai-primitives",
-      "version": "2.6.0",
+      "version": "2.7.0",
       "dependencies": {
         "@anthropic-ai/claude-agent-sdk": "^0.2.119",
         "adf-to-markdown": "^1.0.1",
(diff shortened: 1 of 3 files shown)


rfv-code:
Closing as duplicate of #112 — same Slowloris window in the attachment path, but #112 is the strictly broader fix.

#112 deletes `fetch_attachment.ts`'s local `AbortController`/timer and routes the request through `fetchWithCaps`, which keeps the timeout armed for the whole body **and** enforces the byte cap mid-stream. This PR keeps the local timer and moves the rest of the function inside the timed `try`: the timeout is now armed during `arrayBuffer()`, but the full body is still buffered into memory before the `rawBytes.byteLength > maxBytes` check at `src/toolkit/fetch_attachment.ts`, so the memory-exhaustion path survives an absent or lying `Content-Length` (bounded only by `timeoutMs`).

It also drags a stray `package-lock.json` hunk (2.6.0→2.7.0 and drops the optional `gcp-metadata` peer) — the same artifact that drew a real review objection on earlier regens of this cluster.

**Worth cherry-picking onto #112:** the `tests/toolkit/fetch_attachment.test.ts` Slowloris case. #112's regression test lives in `fetch_helper.test.ts`, so an attachment-level test would be genuinely new coverage.

PR #151, opened 2026-08-02
⚡ Bolt: Optimize SQL statement boundary tracking array

rfv-code:
**💡 What:** 
Replaced `Set<number>` and subsequent `.sort()` calls for gathering statement-terminating semicolon boundaries with a sequentially populated `number[]` array during SQL string evaluation in `src/connectors/db/lib/policy.ts`.

**🎯 Why:**
Tracking index positions during iteration (e.g. tracking boundaries) produces naturally sorted arrays. Placing them into a `Set` required converting them back and running V8's array sort operation, transforming a fast O(N) pass into an unnecessary O(N log N) bottleneck on heavily compound statements.

**📊 Impact:** 
Eliminates overhead for DB policy statement parsing and reduces memory allocation.

**🔬 Measurement:** 
Run `pnpm test tests/connectors/db/policy.test.ts`. Wait time overhead drops on extremely long strings and heavily compound scripts.

---
*PR created automatically by Jules for task [3011801121725844879](https://jules.google.com/task/3011801121725844879) started by @rfv-code*

### DIFF
--- a/src/connectors/db/lib/policy.ts
+++ b/src/connectors/db/lib/policy.ts
@@ -133,20 +133,30 @@ const _DECISION_RANK: Record<Decision, number> = {
 // -----------------------------------------------------------------------
 
 const _READ_KEYWORDS: ReadonlySet<string> = new Set([
-  "SELECT", "EXPLAIN", "SHOW", "DESCRIBE", "DESC", "WITH", "VALUES", "TABLE",
+  "SELECT",
+  "EXPLAIN",
+  "SHOW",
+  "DESCRIBE",
+  "DESC",
+  "WITH",
+  "VALUES",
+  "TABLE",
 ]);
 const _WRITE_KEYWORDS: ReadonlySet<string> = new Set([
-  "INSERT", "UPDATE", "REPLACE", "MERGE", "UPSERT",
+  "INSERT",
+  "UPDATE",
+  "REPLACE",
+  "MERGE",
+  "UPSERT",
 ]);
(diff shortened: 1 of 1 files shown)


rfv-code:
Closing as duplicate of #111.

Same transform, verified hunk-by-hunk against #111 (`faaf9b9e`):
- `_boundarySemicolons` return type `Set<number>` → `number[]`, `new Set<number>()` → `[]`, `boundaries.add(i)` → `boundaries.push(i)`
- `_splitStatements`: `b1.size !== b2.size` → `b1.length !== b2.length`, both `Array.from(...).sort((a,b) => a-b)` calls dropped, loop compares `b1[i]`/`b2[i]` directly, `const boundaries = b1`

Behavior-preserving for the same reason in both: `i` increases monotonically through the scan loop, so the pushed array is already sorted and the `.sort()` was a no-op; `.size` ≡ `.length`; the fail-closed `throw new Error("Ambiguous SQL statement boundaries")` guards are untouched.

#111 is kept because it is the minimal diff (+9/-9). This PR is +108/-47 — the extra ~99 lines are a whole-file Prettier reflow unrelated to the optimization (multi-line conditionals through `_stripComments` / `_dollarQuoteEnd` tag loop, the `scrubbed.length > 500` ternary in `_emitPresentOnly`, and the escape-handling conditionals in `_maskStringLiterals`). That churn makes the security-relevant parser harder to review for no behavioral gain.

Merge-order note for the maintainer: **#94 before #111** — both edit `_splitStatements`.

If this branch covers ground #111 does not, please cherry-pick it onto #111 rather than reopening.

PR #152, opened 2026-08-02
🛡️ Sentinel: [CRITICAL] Fix Slowloris DoS vulnerability in fetch_helper

rfv-code:
🚨 **Severity:** CRITICAL
💡 **Vulnerability:** Uncontrolled Resource Consumption (DoS/Slowloris). The timeout timer in `fetchWithCaps` was cleared immediately after the `fetch` promise resolved (which happens when headers are received). A malicious server could return headers quickly but trickle the response body indefinitely, bypassing the timeout and hanging the `reader.read()` loop.
🎯 **Impact:** An attacker could tie up system connections and exhaust available memory or event loop bandwidth by keeping connections alive indefinitely through slow reading.
🔧 **Fix:** Modified `fetchWithCaps` to enclose both the initial `fetch` request and the entire `reader.read()` stream loop within a single `try...finally` block. This ensures that the timeout remains active until the entire response body is downloaded or the timeout expires.
✅ **Verification:** Verified that tests pass via `npx vitest run tests/toolkit/fetch_helper.test.ts` and confirmed via code review that the fix completely resolves the issue without unnecessary side effects.

---
*PR created automatically by Jules for task [5591141680058406806](https://jules.google.com/task/5591141680058406806) started by @rfv-code*


rfv-code:
Closing as duplicate of #112.

The `fetch_helper.ts` hunk here is the same fix #112 (`f6718d34`) already carries: move the content-length check, the `getReader()` / `read()` drain loop, and the `new Response(merged, ...)` reassembly inside the timed `try`, with `clearTimeout(timer)` relocated to an outer `finally`. That is the correct shape — on `main` today the timer is cleared at `fetch_helper.ts:94`, before the reader loop at `:106`/`:114`, so a stalled body has no active timeout.

#112 is kept because it is a strict superset:
- same `finally` restructure of `fetchWithCaps`
- plus a `fetchImpl` seam so the timeout path is testable
- plus `fetch_attachment.ts` (+3/-20): deletes that file own controller/timer and routes it through `fetchWithCaps`, so attachments get the **mid-stream** byte cap. This PR does not touch `fetch_attachment.ts` at all, leaving that path buffering the whole body before the `maxBytes` check — memory-exhaustion still survives an absent or lying `Content-Length`.
- plus the regression test in `tests/toolkit/fetch_helper.test.ts` (+27), which stalls a `ReadableStream` and asserts the abort fires. This PR ships no test.

If this branch covers ground #112 does not, please cherry-pick onto #112 rather than reopening.

PR #153, opened 2026-08-03
⚡ Bolt: Optimize SQL statement boundaries splitting

rfv-code:
💡 **What**: Changed `_boundarySemicolons` to return `number[]` instead of `Set<number>`, eliminating the need for `Array.from(set).sort()` inside `_splitStatements`.

🎯 **Why**: When iterating over the SQL string, the loop processes indices strictly from left to right. Pushing the semicolon index `i` into an array guarantees the indices are in increasing order. The use of a `Set` required a conversion to array and a redundant O(N log N) `.sort()` operation.

📊 **Impact**: Removes an unnecessary O(N log N) sorting operation on SQL statement splitting, replacing it with an O(1) `.push()` append per delimiter. Reduces overhead and memory allocation (no `Set` or `Array.from()`).

🔬 **Measurement**: Run the full test suite (`pnpm test`) to verify correctness, specifically focusing on `policy.test.ts` where statement splitting occurs.

---
*PR created automatically by Jules for task [6412043930090151444](https://jules.google.com/task/6412043930090151444) started by @rfv-code*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -10,3 +10,6 @@
 ## 2024-06-25 - Avoid array sort for Top-K in hot paths
 **Learning:** V8 engine sorting (`[...arr].sort()`) is surprisingly slow and blocks the event loop for thousands of records when fetching Top-K elements (e.g. usage statistics). It scales at O(N log N) when O(N) is sufficient for a fixed K.
 **Action:** When gathering top elements from a large data array in this codebase, manually track the top items in a single O(N) pass rather than sorting a full copy.
+## 2024-05-27 - Sequential Array Population Avoids Sorting
+**Learning:** Using `Set<number>` to collect monotonically increasing indices from left-to-right string iteration requires a redundant `Array.from(set).sort(...)` step, introducing $O(N \log N)$ sorting overhead where $O(N)$ is achievable.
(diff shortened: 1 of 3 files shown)


rfv-code:
Closing as duplicate of #111.

Both PRs make the identical transform in `src/connectors/db/lib/policy.ts`: `_boundarySemicolons` returns `number[]` instead of `Set<number>` (`.add`→`.push`), and `_splitStatements` drops both `Array.from(...).sort((a,b)=>a-b)` calls, comparing `b1[i]`/`b2[i]` and lengths directly. Diff-verified byte-equivalent in the core hunks; the only source-level delta is variable naming (#153 keeps `b1Array`/`b2Array`, #111 renames to `b1`/`b2` and adds a rationale comment on why the monotonic push makes the sort a no-op).

#111 is kept because it carries that rationale comment and — unlike this PR — does **not** drag an unrelated `package-lock.json` hunk. This PR's lockfile change bumps 2.6.0→2.7.0 and drops the root `gcp-metadata@7.0.1` entry, which `mongodb` declares as an optional peer for `MONGODB-GCP` auth; the same hunk has drawn a real Codex P2 on prior PRs in this cluster.

Note for the maintainer: merge **#94 before #111** — both edit `_splitStatements`.

If this PR covers ground #111 doesn't, please cherry-pick onto #111 rather than reopening.

PR #157, opened 2026-08-05
🛡️ Sentinel: [HIGH] Fix Slowloris DoS vulnerability in fetch_helper

rfv-code:
🚨 **Severity**: HIGH
💡 **Vulnerability**: A Slowloris-style DoS vulnerability existed in `fetchWithCaps` (`src/toolkit/fetch_helper.ts`). The abort controller timeout was cleared immediately after the `fetch` promise resolved (which happens when HTTP headers are received), leaving the stream consumption phase unprotected against a server that maliciously trickles chunks of the response body.
🎯 **Impact**: An attacker or misconfigured upstream service could indefinitely tie up resources and block the Node.js event loop by sending an endless, extremely slow response body stream.
🔧 **Fix**: The entire stream consumption logic is now wrapped inside a `try...finally` block alongside the `fetch` call, and the `clearTimeout(timer)` call was moved to the `finally` block to ensure the request is aborted if it exceeds `timeoutMs` during any phase of execution.
✅ **Verification**: Verified using a custom node script demonstrating a slow streaming response (`setInterval` chunk trickle), confirming the script now successfully errors out with an abort timeout instead of hanging. Also ran `pnpm test tests/toolkit/fetch_helper.test.ts` to ensure normal functionality remains completely unimpeded without regressions. Cleaned up all manual testing/lock artifacts.

---
*PR created automatically by Jules for task [4666241091890670036](https://jules.google.com/task/4666241091890670036) started by @rfv-code*


rfv-code:
Closing as duplicate of #112, which is a strict superset of this change.

The `fetch_helper.ts` fix here is the same one #112 makes: move the content-length check, the `getReader().read()` drain loop, and the `new Response(merged)` reassembly inside the timed `try`, and relocate `clearTimeout(timer)` to an outer `finally` so the abort timer stays armed through body streaming. That diagnosis is correct — on `main` the timer is cleared at `fetch_helper.ts:94`, before `getReader()` at `:106` and `read()` at `:114`, so a server that trickles the body is unbounded.

#112 does everything this PR does, and three things it does not:

1. **`fetch_attachment.ts` is routed through `fetchWithCaps`** (+3/-20), deleting that file's own controller/timer. This PR leaves `fetch_attachment.ts` untouched, so attachment fetches still buffer the whole body via `arrayBuffer()` before the `rawBytes.byteLength > maxBytes` check — memory exhaustion survives an absent or lying `Content-Length`, bounded only by `timeoutMs`.
2. **A regression test** (`tests/toolkit/fetch_helper.test.ts`, +27) driving a stalled `ReadableStream`, plus the `fetchImpl` seam that makes it injectable. This PR ships no test, so nothing pins the fix against a later regeneration.
3. **The abort-reason fix (`c04ec6f`).** `fetch_helper.ts` documents that a timeout surfaces as a `DOMException` named `AbortError`, but the timer aborts with `new Error("fetch_helper timeout")` and the abort reason propagates verbatim to whatever `fetch`/`read()` rejects with — so `err.name` is `"Error"`, which is the exact check at `github_client.ts:662` and in the gitlab/linear/notion/jira/confluence suites. This PR keeps `timeoutCtl.abort(new Error("fetch_helper timeout"))` unchanged at line 95, and because it widens the timed window to cover body streaming, it widens the mis-typed-abort exposure along with it.

One thing this PR does better: it is clean — no stray `package-lock.json` hunk, which several PRs in this cluster have carried.

If it covers ground #112 does not, please cherry-pick onto #112 rather than reopening.

PR #161, opened 2026-08-07
⚡ Bolt: Replace Set with Array for sequential index tracking

rfv-code:
💡 **What**: Replaced `Set<number>` with a standard `number[]` array for storing statement boundaries in `_boundarySemicolons`, removing the subsequent `Array.from().sort()` conversions in `_splitStatements`.
🎯 **Why**: When iterating sequentially through a string and recording indices, the indices are naturally monotonic. Collecting them in a `Set` and then sorting them introduces unnecessary `O(N log N)` overhead and memory allocations.
📊 **Impact**: Reduces the constant factor and algorithmic overhead `O(N log N)` -> `O(N)` during statement boundary tracking for every executed SQL query passing through the policy engine.
🔬 **Measurement**: Verified that all unit tests (`tests/connectors/db/policy.test.ts`) pass and the logic is identical by reviewing the resulting array properties.

---
*PR created automatically by Jules for task [12216786512630502994](https://jules.google.com/task/12216786512630502994) started by @rfv-code*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -10,3 +10,6 @@
 ## 2024-06-25 - Avoid array sort for Top-K in hot paths
 **Learning:** V8 engine sorting (`[...arr].sort()`) is surprisingly slow and blocks the event loop for thousands of records when fetching Top-K elements (e.g. usage statistics). It scales at O(N log N) when O(N) is sufficient for a fixed K.
 **Action:** When gathering top elements from a large data array in this codebase, manually track the top items in a single O(N) pass rather than sorting a full copy.
+## 2026-05-26 - Sequential Index Tracking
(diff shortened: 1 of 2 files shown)


rfv-code:
Closing as duplicate of #111.

Diff-verified byte-identical to #163 and #165, and semantically identical to #111: `_boundarySemicolons` `Set<number>` to `number[]`, `.add` to `.push`, `.size` to `.length`, both `Array.from(...).sort()` calls dropped from `_splitStatements`, `b1[i]`/`b2[i]` compared directly. Remaining delta against #111 is a whole-file prettier reflow.

Three regenerations of this same optimization landed in three days (#161, #163, #165). #111 remains the canonical: minimal diff, plus a comment recording why removing the sort is safe.

Note for merge ordering: **#94 should merge before #111** — both edit `_splitStatements`.

PR #162, opened 2026-08-07
🛡️ Sentinel: [CRITICAL] Fix Slowloris DoS in fetchWithCaps

rfv-code:
🚨 **Severity:** CRITICAL
💡 **Vulnerability:** Slowloris DoS vulnerability in `fetchWithCaps` where the timeout was cleared after headers resolved, rather than after the body was fully read.
🎯 **Impact:** An attacker or misbehaving server could intentionally trickle the response body over an open connection indefinitely. Because the timeout was already cleared, the application would block/hang resources until the (potentially unbounded) stream either naturally closed or memory ran out (which the bytes cap might not catch if trickled slowly enough).
🔧 **Fix:** Wrapped the response body `reader.read()` loop inside a `try...finally` block, moving the `clearTimeout(timer)` call into the `finally` clause so the timer protects the entire request lifecycle.
✅ **Verification:** Verified by running `pnpm test tests/toolkit/fetch_helper.test.ts`.

---
*PR created automatically by Jules for task [1747153796188009003](https://jules.google.com/task/1747153796188009003) started by @rfv-code*


rfv-code:
Closing as duplicate of #112.

`fetch_helper.ts` here is the same finally-wrap #112 already carries: the content-length check, the `getReader().read()` drain loop and the `new Response(merged)` reassembly all move inside the timed `try`, with `clearTimeout` in an outer `finally`. That is the correct fix for the Slowloris window, and #112 has it.

#112 is a strict superset. It additionally has the `fetchImpl` seam, routes `fetch_attachment.ts` through `fetchWithCaps` so the cap applies mid-stream instead of after a full buffer, ships an 85-line regression test, and aborts with `new DOMException("fetch_helper timeout", "AbortError")` rather than a plain `Error` — which matters because callers such as `github_client.ts:662` classify cancellation with `err.name === "AbortError"`. This branch keeps the plain `Error`.

This branch also drags the recurring stray `package-lock.json` hunk (+2/-18) that bumps a version and drops the optional `gcp-metadata` peer. That hunk is unrelated to the fix and has drawn a real codex P2 on earlier regenerations. Extra reason not to take this branch.

PR #163, opened 2026-08-08
⚡ Bolt: Replace Set and sort with sequential array in SQL boundary tracking

rfv-code:
💡 What:
Replaced the `Set` with a standard `number[]` array for collecting SQL statement boundaries in `_boundarySemicolons`. Removed the redundant `Array.from().sort()` call in `_splitStatements`.

🎯 Why:
The index variable `i` naturally advances forward during the string parsing loop, meaning discovered indices are inherently unique and pre-sorted. Using a `Set` and subsequently sorting the array introduces unnecessary O(N log N) time complexity and object creation overhead.

📊 Impact:
Removes redundant O(N log N) sorting overhead and reduces memory allocation (no `Set` or `Array.from()` conversions) on the hot path for splitting SQL statements.

🔬 Measurement:
Run `pnpm test tests/connectors/db/policy.test.ts` to ensure policy rules and statement splitting still function identically.

---
*PR created automatically by Jules for task [8801890190008737738](https://jules.google.com/task/8801890190008737738) started by @rfv-code*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -10,3 +10,6 @@
 ## 2024-06-25 - Avoid array sort for Top-K in hot paths
 **Learning:** V8 engine sorting (`[...arr].sort()`) is surprisingly slow and blocks the event loop for thousands of records when fetching Top-K elements (e.g. usage statistics). It scales at O(N log N) when O(N) is sufficient for a fixed K.
 **Action:** When gathering top elements from a large data array in this codebase, manually track the top items in a single O(N) pass rather than sorting a full copy.
+## 2026-05-25 - Avoid Set and sort for monotonically increasing indices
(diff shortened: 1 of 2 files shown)


rfv-code:
Closing as duplicate of #111.

Diff-verified byte-identical to #165 and semantically identical to #111: `_boundarySemicolons` `Set<number>` to `number[]`, `.add` to `.push`, `.size` to `.length`, both `Array.from(...).sort()` calls dropped from `_splitStatements`, `b1[i]`/`b2[i]` compared directly. The only delta against #111 is a whole-file prettier reflow.

#111 wins on being the minimal diff (+9/-9) and on documenting why the sort was removable — boundaries are pushed in monotonic `i` order, so the sort was a no-op.

Note for merge ordering: **#94 should merge before #111**, since both edit `_splitStatements`.

PR #164, opened 2026-08-08
🛡️ Sentinel: [HIGH] Fix DoS vulnerabilities in fetch timeouts and unbounded memory consumption

rfv-code:
🚨 **Severity**: HIGH
💡 **Vulnerability**: In `src/toolkit/fetch_helper.ts` and `src/toolkit/fetch_attachment.ts`, the timeout `clearTimeout()` was called immediately after the initial HTTP headers were resolved. The subsequent body reading loops (`reader.read()` and `response.arrayBuffer()`) had no timeout enforced. Furthermore, `fetchAttachment` accumulated bytes in memory entirely without bounds checking during download.
🎯 **Impact**: A malicious server could intentionally trickle a response body very slowly (Slowloris attack), permanently hanging the connector process and tying up system resources. Additionally, large responses could cause DoS memory exhaustion before the final manual check.
🔧 **Fix**: 
1. Moved the entire body reading loop inside a `try..finally` block in `fetchWithCaps` to ensure the `clearTimeout` call is only made *after* stream processing finishes. 
2. Updated `fetchAttachment` to completely rely on `fetchWithCaps`, taking advantage of the secured bounded streaming functionality and dropping the insecure `arrayBuffer()` fallback. Added `fetchImpl` to options to keep tests working seamlessly.
✅ **Verification**: Run `pnpm test tests/toolkit/fetch_helper.test.ts tests/toolkit/fetch_attachment.test.ts` to ensure functionality and tests still pass as intended.

---
*PR created automatically by Jules for task [5432998316479657332](https://jules.google.com/task/5432998316479657332) started by @rfv-code*


rfv-code:
Closing as duplicate of #112.

This is the closest any regeneration in this cluster has come to #112 — it is the first to carry both halves of the fix (the timed-`try` wrap in `fetchWithCaps` **and** routing `fetch_attachment.ts` through `fetchWithCaps` so the byte cap applies mid-stream rather than after a full buffer). Verified against #112's tip, though, it is still a strict subset on two points:

1. **No regression test.** #112 ships `tests/toolkit/fetch_helper.test.ts` +85, including a stalled-`ReadableStream` case that hangs on pre-fix code. This branch adds none.
2. **It lacks the `AbortError` fix and widens the bug it omits.** This branch keeps `timeoutCtl.abort(new Error("fetch_helper timeout"))`. The abort reason propagates verbatim to whatever `fetch`/`reader.read()` rejects with, so `err.name` is `"Error"` — while the docblocks at `fetch_helper.ts:4-6` and `:73-74` advertise `DOMException`/`AbortError`, and callers such as `github_client.ts:662` classify cancellation with `err.name === "AbortError"`. #112 fixes this in `c04ec6f` by aborting with `new DOMException("fetch_helper timeout", "AbortError")`, with three tests pinning it. Because this branch keeps the timer armed through the body stream, it *widens* the window through which the mis-typed abort escapes.

One genuine difference worth recording, not a defect: this branch threads `fetchImpl` as a field on the caps options object, where #112 passes it as a fourth positional parameter. The options form is arguably the cleaner API. It is a small enough delta that it is not worth reopening the merged shape for, but if the maintainer prefers it, it cherry-picks cleanly onto #112.

PR #167, opened 2026-08-10
⚡ Bolt: Remove O(N log N) sorting overhead in SQL statement splitting

rfv-code:
💡 What: Refactored `_boundarySemicolons` in `src/connectors/db/lib/policy.ts` to return `number[]` instead of `Set<number>`, populating it with `.push()` instead of `.add()`. Updated `_splitStatements` to directly use these naturally ordered arrays instead of converting them from sets and sorting them.

🎯 Why: The previous implementation used a `Set` to collect semicolon boundaries, then converted it to an array and sorted it with `.sort()`. However, because the string is iterated sequentially, the boundaries are discovered in strictly increasing order. The `.sort()` step is entirely redundant and introduces an unnecessary O(N log N) performance overhead for large SQL statements.

📊 Impact: Eliminates O(N log N) sorting step when parsing SQL statements in the policy engine, directly dropping the overhead to O(1) for insertion and O(N) overall.

🔬 Measurement: Verified with `pnpm test tests/connectors/db/policy.test.ts` to ensure no changes to split behavior.

---
*PR created automatically by Jules for task [386113578415658636](https://jules.google.com/task/386113578415658636) started by @rfv-code*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -10,3 +10,4 @@
 ## 2024-06-25 - Avoid array sort for Top-K in hot paths
 **Learning:** V8 engine sorting (`[...arr].sort()`) is surprisingly slow and blocks the event loop for thousands of records when fetching Top-K elements (e.g. usage statistics). It scales at O(N log N) when O(N) is sufficient for a fixed K.
 **Action:** When gathering top elements from a large data array in this codebase, manually track the top items in a single O(N) pass rather than sorting a full copy.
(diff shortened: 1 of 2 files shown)


rfv-code:
Closing as duplicate of #111.

Same transform, verified hunk by hunk: `_boundarySemicolons` returns `number[]` instead of `Set<number>`, `boundaries.add(i)` becomes `boundaries.push(i)`, `b1.size !== b2.size` becomes `b1.length !== b2.length`, both `Array.from(...).sort((a, b) => a - b)` calls are dropped, and the comparison loop indexes `b1[i]`/`b2[i]` directly. #111 is byte-identical on those lines.

#111 is kept because it is minimal — `+9/-9`, one file, no `.jules/bolt.md` churn — while this PR reflows the whole file (`+112/-47`), including collapsing `_DELETE_KEYWORDS` and `_PRIVILEGE_KEYWORDS` to single lines. #111 also carries a two-line rationale comment recording *why* the sort is safe to drop (the boundaries are pushed in monotonic order), which is the part a future reader needs.

**Merge order note for the maintainer:** merge #94 before #111. Both edit `_splitStatements`.

If this PR covers ground #111 does not, please cherry-pick it onto #111.

PR #171, opened 2026-08-12
⚡ Bolt: Remove Set and sort overhead in SQL boundary tracking

rfv-code:
💡 What: Refactored `_boundarySemicolons` in `src/connectors/db/lib/policy.ts` to push sequential indices into a standard array instead of accumulating them in a `Set`. This removes the `Array.from(set).sort()` overhead in `_splitStatements`.

🎯 Why: The parsing loop naturally iterates over the string sequentially (`let i = 0; i < sql.length; i++`). This guarantees that any discovered boundaries are inherently unique and already in ascending order. Using a `Set` followed by a sort introduces unnecessary allocations and an O(N log N) algorithmic bottleneck on large SQL statements.

📊 Impact: Reduces O(N log N) sorting overhead to O(1) appending per boundary, which speeds up compound statement evaluation under load.

🔬 Measurement: Verify by running `pnpm test tests/connectors/db/policy.test.ts`. All 59 tests should pass, confirming behavior remains completely identical.

---
*PR created automatically by Jules for task [17213055597281726997](https://jules.google.com/task/17213055597281726997) started by @rfv-code*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -10,3 +10,4 @@
 ## 2024-06-25 - Avoid array sort for Top-K in hot paths
 **Learning:** V8 engine sorting (`[...arr].sort()`) is surprisingly slow and blocks the event loop for thousands of records when fetching Top-K elements (e.g. usage statistics). It scales at O(N log N) when O(N) is sufficient for a fixed K.
 **Action:** When gathering top elements from a large data array in this codebase, manually track the top items in a single O(N) pass rather than sorting a full copy.
+## 2026-08-12 - Remove Set and sort overhead in SQL boundary tracking\n**Learning:** When parsing strings to find boundary indices, sequentially pushing to a number array is faster and automatically sorted.\n**Action:** Avoid using `Set` and `Array.from(set).sort()` for simple sequential index tracking to prevent O(N log N) performance bottlenecks.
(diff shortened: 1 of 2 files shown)


rfv-code:
Closing as duplicate of #111, which implements the same optimization.

**Verified identical transform.** Both PRs change `_boundarySemicolons` to return `number[]` instead of `Set<number>`, swap `boundaries.add(i)` for `boundaries.push(i)`, compare `.length` instead of `.size`, and drop both `Array.from(...).sort((a, b) => a - b)` calls in `_splitStatements`. The core hunks match line for line.

**The only source delta is variable naming.** This PR renames the declarations to `b1Array` / `b2Array`; #111 uses `b1` / `b2` at every site and adds a rationale comment recording *why* dropping the sort is safe — `_boundarySemicolons` scans `sql` left to right and pushes each boundary as it finds it, so the array is already in monotonic index order. That invariant is the load-bearing part of this change. If a later edit makes the scanner emit out of order, the element-wise comparison loop silently mis-splits statements instead of throwing "Ambiguous SQL statement boundaries", so #111 keeps the comment that pins it.

**Credit where it is due: this is the cleanest regeneration this cluster has produced.** A bare +6/-8 on `policy.ts`, no whole-file prettier reflow, and no `package-lock.json` or `pnpm-lock.yaml` churn — all three of which have shipped on prior PRs in this series and each cost a separate review cycle.

If this PR covers ground #111 does not, please cherry-pick onto #111 rather than reopening.

PR #172, opened 2026-08-13
🛡️ Sentinel: [CRITICAL] Fix Slowloris DoS in fetch_helper

rfv-code:
🚨 **Severity:** CRITICAL
💡 **Vulnerability:** The `fetch_helper` utility cleared its timeout immediately after receiving the HTTP headers, before the body was fully read. This left the application vulnerable to a Slowloris-style Denial of Service (DoS) attack, where a malicious server could stream the response body indefinitely, consuming worker threads or memory. Furthermore, `fetch_attachment` was calling `arrayBuffer()` on responses directly before size checking, which risks Out-of-Memory (OOM) crashes if a malicious server sends a massive payload without a `content-length` header.
🎯 **Impact:** An attacker could exhaust server resources by tying up connections and threads indefinitely, leading to service degradation or a full denial of service. They could also cause the node process to crash via an OOM error.
🔧 **Fix:** Refactored `fetchWithCaps` in `src/toolkit/fetch_helper.ts` to clear the timeout only *after* the entire body streaming operation completes (in a `finally` block). Also refactored `fetch_attachment.ts` to utilize the fortified `fetchWithCaps` instead of reading the `arrayBuffer` unsafely, ensuring chunk-by-chunk stream length monitoring.
✅ **Verification:** Verified by unit tests in `fetch_helper.test.ts` and `fetch_attachment.test.ts` to ensure memory limits and timeouts still function effectively. Checked the `finally` block positioning manually to confirm timer retention during stream processing.

---
*PR created automatically by Jules for task [12304565632927127474](https://jules.google.com/task/12304565632927127474) started by @rfv-code*


rfv-code:
Closing as duplicate of #112, which is the standing canonical for the Slowloris / fetch-caps cluster.

**This PR gets further than most regenerations in this cluster — it carries both halves.** It has the finally-wrap (`clearTimeout` at `fetch_helper.ts:137`, after `getReader()` `:105` and `read()` `:113`, so the streaming loop really is inside the timed `try`), **and** it routes `fetch_attachment.ts` through `fetchWithCaps` (`:99`) instead of buffering before the cap check. That is the fifth time this cluster has produced the pair, and it is the right shape.

**But it is still a strict subset of #112 on two axes, and net-negative on one of them.**

**1. It keeps the mis-typed abort reason — while widening the window that reason escapes through.**

At `:88` this PR retains:

```ts
const timer = setTimeout(() => timeoutCtl.abort(new Error("fetch_helper timeout")), timeoutMs);
```

Its own doc comment at `:75` promises `DOMException` / `AbortError` on timeout. The abort reason propagates verbatim to whatever `fetch`/`reader.read()` rejects with, so callers checking `err.name === "AbortError"` — which is what `github_client.ts:662` and the gitlab/linear/notion/jira/confluence suites do — see `err.name === "Error"` and misclassify the timeout.

This is pre-existing on `main` (`:87`), so the PR does not introduce it. The problem is the interaction: by keeping the timer armed through the body stream, this PR makes timeouts fire on a mid-drain path that previously could not time out at all. Merging it alone therefore pushes the misclassification onto **more** callers than `main` has today. The regression is not "missing a fix" — on that axis it is worse than the status quo.

#112 fixed this in `c04ec6f`, and its tip still carries it:

```ts
// fetch_helper.ts:95
() => timeoutCtl.abort(new DOMException("fetch_helper timeout", "AbortError")),
```

**2. No tests.** This PR is +51/-58 across three files with no test changes. #112 ships `tests/toolkit/fetch_helper.test.ts` **+85**, covering the header phase, the body-stream phase, and one case pinning that caller-supplied abort reasons stay untouched so the fix cannot over-correct.

**One real difference that is arguably better, and is not noise.** This PR threads the test seam as an options-object field (`caps.fetchImpl`, declared `:27`, used `:93`) where #112 uses a 4th positional parameter (`:82`, `:105`). The options form is cleaner and this cluster has now proposed it many times. It has never been adopted, and it is not worth a re-review of the canonical on its own — but if you prefer that signature, cherry-pick just that hunk onto #112.

Premises re-verified still unlanded on `main` (`fbd912a1`) before closing: `fetch_helper.ts:87` plain-`Error` abort, `:94` `clearTimeout` before the body phase, `getReader()` `:106`, `read()` `:114`.

PR #173, opened 2026-08-13
⚡ Bolt: Replace Set and sort for SQL boundaries with simple Array

rfv-code:
💡 **What:** 
Replaced `Set<number>` with `number[]` in `_boundarySemicolons` inside `policy.ts`, and removed the `Array.from(set).sort()` overhead in `_splitStatements`.

🎯 **Why:** 
When parsing SQL and finding semicolon boundaries, the string index strictly moves left-to-right (monotonically increasing). Using a `Set` was unnecessary because indices are intrinsically unique, and sorting the set elements later scaled at O(N log N).

📊 **Impact:** 
Changes a previously O(N log N) array sort (plus Set allocation overhead) into a clean O(N) array append. While N (semicolons) is typically small, this optimization avoids needless allocation and sorting inside the hot-path policy engine's dual-mode parsing.

🔬 **Measurement:** 
Unit tests pass successfully. The logic is logically equivalent since the loop naturally populates elements sequentially.

---
*PR created automatically by Jules for task [5540761225263123375](https://jules.google.com/task/5540761225263123375) started by @rfv-code*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -10,3 +10,6 @@
 ## 2024-06-25 - Avoid array sort for Top-K in hot paths
 **Learning:** V8 engine sorting (`[...arr].sort()`) is surprisingly slow and blocks the event loop for thousands of records when fetching Top-K elements (e.g. usage statistics). It scales at O(N log N) when O(N) is sufficient for a fixed K.
 **Action:** When gathering top elements from a large data array in this codebase, manually track the top items in a single O(N) pass rather than sorting a full copy.
+## 2024-07-28 - Set to Array for Sequential Loops
+**Learning:** Using `Set` to collect sequentially increasing indices and then later converting to an array and sorting (`Array.from(set).sort()`) introduces completely unnecessary memory allocation and an O(N log N) performance hit.
+**Action:** When tracking index positions sequentially in string or array processing (like SQL parsing), push directly to a standard array. The output is naturally sorted and unique.
--- a/src/connectors/db/lib/policy.ts
+++ b/src/connectors/db/lib/policy.ts
@@ -133,20 +133,30 @@ const _DECISION_RANK: Record<Decision, number> = {
 // -----------------------------------------------------------------------
 
 const _READ_KEYWORDS: ReadonlySet<string> = new Set([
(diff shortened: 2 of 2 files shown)


rfv-code:
Closing as duplicate of #111, which is the standing canonical for the SQL-boundary `Set`/`sort` cluster.

**The core transform is identical.** Diffed both branches against `main` (`fbd912a1`) with `-w --ignore-blank-lines`; the load-bearing hunks in `src/connectors/db/lib/policy.ts` match line for line:

- `_boundarySemicolons` return type `Set<number>` → `number[]`
- `new Set<number>()` → `[]`, `.add(i)` → `.push(i)`
- both `Array.from(...).sort((a, b) => a - b)` calls dropped
- comparison reads `b1[i]` / `b2[i]`, and `const boundaries = b1`

Neither branch leaves a dead `const b1Array = b1;` alias behind.

**Why #111 is kept.** #111 is **+9/-47 → +9/-9**, exactly the transform and nothing else. This PR is **+115/-47**, and the entire remaining delta is a whole-file prettier reflow that is unrelated to the optimization.

I measured the reflow direction rather than assuming it, because it has gone both ways on this cluster:

| | lines | lines over 80 cols |
|---|---|---|
| `main` | 1006 | 31 |
| #111 | 1006 | 33 |
| this PR | 1071 | 7 |

So #111 sits at `main`'s prevailing formatting, while this PR reformats `policy.ts` to a **narrower** print width than the file is currently written in — a reflow `main` has not adopted. That is churn in a security-relevant parser, and it makes the one real change hard to see in review.

One genuine merit worth recording: this PR comments **both** sites (the `boundaries` declaration and the comparison loop), where #111 comments only the loop. #111's single comment already pins the invariant that makes dropping the sort safe — that indices are pushed in monotonically increasing order — so the coverage difference is presentational, not substantive.

Premise re-verified still unlanded on `main` before closing: `policy.ts:254` `Set<number>`, `:255`, `:283` `.add`, `:309` `.size`, `:312`/`:313` `Array.from().sort()`.

If you want the second comment or any of the reflow, please cherry-pick onto #111.

PR #175, opened 2026-08-14

rfv-code:
💡 **What:** 
Replaced `Set<number>` with a standard `number[]` array in the `_boundarySemicolons` function, and removed the subsequent `Array.from(...).sort(...)` calls in `_splitStatements`.

🎯 **Why:** 
When iterating over a SQL string sequentially (`for (let i = 0; i < sql.length; i++)`), the matched boundary indices are naturally distinct and monotonically increasing. Storing them in a `Set` and then converting to an array to sort it introduces unnecessary object allocations and a redundant $O(N \log N)$ sorting step for data that is already strictly sorted in $O(N)$ time. 

📊 **Impact:** 
Eliminates unnecessary $O(N \log N)$ sorting and Set-to-Array object allocations during SQL statement parsing in the DB connector policy engine, slightly improving throughput for policy evaluation on large, compound SQL scripts without sacrificing any readability.

🔬 **Measurement:** 
Run `pnpm test tests/connectors/db/policy.test.ts` to verify the boundary logic remains exactly correct and passes all existing policy assertions.

---
*PR created automatically by Jules for task [11414394881458612655](https://jules.google.com/task/11414394881458612655) started by @rfv-code*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -10,3 +10,6 @@
 ## 2024-06-25 - Avoid array sort for Top-K in hot paths
 **Learning:** V8 engine sorting (`[...arr].sort()`) is surprisingly slow and blocks the event loop for thousands of records when fetching Top-K elements (e.g. usage statistics). It scales at O(N log N) when O(N) is sufficient for a fixed K.
 **Action:** When gathering top elements from a large data array in this codebase, manually track the top items in a single O(N) pass rather than sorting a full copy.
+## 2024-05-15 - Array collection for monotonically increasing indices
+**Learning:** Collecting indices during a sequential string/array pass (e.g. finding statement boundaries) naturally results in a sorted list with unique items. Using a `Set` and subsequently converting to an Array and sorting it (`Array.from(set).sort()`) introduces unnecessary object allocation and O(N log N) sorting overhead for something that is already O(N) and sorted.
+**Action:** When gathering positions/indices during a forward pass over a string or array, use standard arrays (`number[]`) and `.push(i)` to avoid redundant post-processing overhead.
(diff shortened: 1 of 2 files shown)


rfv-code:
Closing as duplicate of #111.

Both PRs make the identical transform in `src/connectors/db/lib/policy.ts`: `_boundarySemicolons` returns `number[]` instead of `Set<number>`, `.add(i)` becomes `.push(i)`, `.size` becomes `.length`, and both `Array.from(...).sort((a, b) => a - b)` calls are dropped from `_splitStatements`. Verified hunk-by-hunk against #111's tip `faaf9b9e`.

This is the cleanest shape this cluster has produced — a bare +7/-8 transform with **no whole-file prettier reflow and no lockfile churn**, and no dead `const b1Array = b1;` aliases. The only substantive delta from #111 is naming and comment placement: this PR renames the declarations to `b1Array`/`b2Array` and comments the `boundaries` declaration inside `_boundarySemicolons`; #111 keeps `b1`/`b2` with `const boundaries = b1` and comments the comparison loop in `_splitStatements`.

#111 is kept because it is the standing canonical for this cluster, carries the long-form review, and its comment sits on the comparison loop — the place that actually depends on the monotonic-push invariant that makes dropping the sort safe. Both are equally minimal, so this is a coin-flip on presentation, not on substance.

Premise re-verified unlanded on `main`: `policy.ts:254` `Set<number>`, `:255`, `:283` `.add`, `:309` `.size`, `:312`/`:313` `Array.from().sort()`.

If you consider the declaration-site comment or the `b1Array`/`b2Array` naming an improvement, please cherry-pick onto #111.

PR #176, opened 2026-08-15
⚡ Bolt: Remove redundant sort and Set allocation in boundary tracking

rfv-code:
💡 **What:** Replaced `Set<number>` with `number[]` to collect semicolon boundary indexes during sequential SQL parsing. Because the string is processed left-to-right, indexes naturally append in monotonically increasing order, making `Set` insertion and subsequent sorting redundant.
🎯 **Why:** To eliminate unnecessary O(N) allocation of a `Set`, O(N) array conversion, and O(N log N) array sorting overhead in `_splitStatements` every time a query is evaluated by the policy engine.
📊 **Impact:** Reduces time complexity for boundary sorting from O(N log N) to O(1) by removing the step entirely, reducing garbage collection pressure and speeding up policy checks on large compound statements.
🔬 **Measurement:** Verified via Vitest (`pnpm test tests/connectors/db/policy.test.ts`), which passes instantly.

---
*PR created automatically by Jules for task [14996889039220027532](https://jules.google.com/task/14996889039220027532) started by @rfv-code*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -10,3 +10,6 @@
 ## 2024-06-25 - Avoid array sort for Top-K in hot paths
 **Learning:** V8 engine sorting (`[...arr].sort()`) is surprisingly slow and blocks the event loop for thousands of records when fetching Top-K elements (e.g. usage statistics). It scales at O(N log N) when O(N) is sufficient for a fixed K.
 **Action:** When gathering top elements from a large data array in this codebase, manually track the top items in a single O(N) pass rather than sorting a full copy.
+## 2024-05-18 - Optimize boundary tracking in string parsing
+**Learning:** When scanning a string left-to-right to collect boundary indexes, the indexes are naturally discovered in monotonically increasing order. Using a `Set` to track these requires later conversion to an array and redundant sorting, introducing unnecessary O(N log N) overhead.
+**Action:** When tracking index positions during sequential string or array iteration, directly use an array (`number[]`) populated with `push()` instead of `Set`s, avoiding redundant sorting.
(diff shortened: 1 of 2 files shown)


rfv-code:
Closing as duplicate of #111.

Both PRs make the same change to `_boundarySemicolons` / `_splitStatements` in `src/connectors/db/lib/policy.ts`: `Set<number>` → `number[]`, `.add` → `.push`, `.size` → `.length`, and both `Array.from().sort()` calls dropped in favour of indexing `b1[i]` / `b2[i]`.

`git diff pr/111 pr/176 -- src/connectors/db/lib/policy.ts` is **two lines**, and they are the only difference between the two branches:

```
-  // ⚡ Bolt Optimization: `_boundarySemicolons` returns arrays populated in monotonic order.
-  // This avoids `O(M log M)` sorting and `Set` allocations overhead on the hot parsing path.
```

That is, this PR is #111 minus #111's rationale comment. Kept #111 because that comment pins the **monotonic-push invariant** — the property that makes dropping the sort safe in the first place. `_boundarySemicolons` walks `sql` with a single increasing `i`, so semicolon indices are pushed in ascending order and the arrays are already sorted; without that note the next reader has to re-derive why removing `.sort()` is not a behaviour change on a security-relevant SQL boundary parser.

Credit where due: this is the **cleanest shape this cluster has produced**. Bare transform, no whole-file prettier reflow, no `package-lock.json` / `pnpm-lock.yaml` churn, and no dead `b1Array` / `b2Array` aliases (cf. #168). Nothing here needs cherry-picking onto #111 — #111 is a strict superset.

Premise re-verified unlanded: `main` (`fbd912a1`) still has `Set<number>` at `policy.ts:254`, `boundaries.add(i)` at `:283`, and `Array.from(b1).sort(...)` at `:312`.

PR #177, opened 2026-08-15
🛡️ Sentinel: [HIGH] Fix fetch stream resource exhaustion (Slowloris) in fetchWithCaps

rfv-code:
🚨 **Severity:** HIGH
💡 **Vulnerability:** Slowloris / Resource Exhaustion DoS in `fetchWithCaps`. The standard `fetch` only retrieves the response headers. Because the `clearTimeout` was executing immediately after the headers arrived, the subsequent `reader.read()` streaming loop was completely unprotected by the timeout. A malicious server could return headers and then stream the body byte-by-byte indefinitely, hanging the connector.
🎯 **Impact:** An attacker could exhaust the system's memory and connections by submitting URLs to endpoints that intentionally stall during body streaming.
🔧 **Fix:** Wrapped the entirety of the fetch header request and the `reader.read()` stream loops within a single `try...finally` block, moving the `clearTimeout(timer)` call to this final block so the timeout properly triggers if the stream hangs.
✅ **Verification:** Verified by generating a vulnerable node.js server using a local script `test_slowloris2.ts` that trickles chunks every 50ms, and validating the timeout successfully fires instead of hanging. Also verified using the `pnpm test` vitest suite.

---
*PR created automatically by Jules for task [3248742259245541613](https://jules.google.com/task/3248742259245541613) started by @rfv-code*


rfv-code:
Closing as duplicate of #112.

Same core fix: keep the streaming body read inside the timed `try` so `clearTimeout` only fires in the `finally` after the body is drained, closing the Slowloris window where a server sends headers promptly then drips the body forever.

#112 is a strict superset, and this PR is **net-negative on one axis** — worth stating precisely rather than filing it as "just a dup":

**1. It reverts the `AbortError` fix while widening the window that bug escapes through.** This PR keeps

```js
() => timeoutCtl.abort(new Error("fetch_helper timeout")),   // :96
```

while its own docblock at `:82` promises `DOMException` / `AbortError`. #112 carries `c04ec6f`, which aborts with `new DOMException("fetch_helper timeout", "AbortError")` (`:95`). Callers doing the conventional `err.name === "AbortError"` check misclassify a timeout as a generic network failure. Because the finally-wrap keeps the timer armed *through the body stream*, the fix makes that misclassification reachable on a **second** path — so shipping the wrap without the `DOMException` change pushes the wrong error type onto more callers than `main` has today.

This is not a hypothetical reading of the interaction: #112's diff contains a comment explaining exactly this coupling, and this PR deletes it —

```
-  // Abort with a DOMException named "AbortError" — the reason propagates
-  // verbatim to whatever `fetch`/`reader.read()` rejects with, and both
-  // docblocks above advertise `AbortError`. A plain `Error` here left callers
-  // doing the conventional `err.name === "AbortError"` check misclassifying a
-  // timeout as a generic network failure. Now that the timer stays armed
-  // through the body stream, that misclassification reaches a second path.
```

**2. Missing scope.** This PR touches `fetch_helper.ts` only. #112 also routes `src/toolkit/fetch_attachment.ts` through `fetchWithCaps` (`:99`), which is what gets `fetchAttachment` a **mid-stream** byte cap instead of buffering first and checking after.

**3. No test.** #112 ships +85 lines in `tests/toolkit/fetch_helper.test.ts`, including the stalled-`ReadableStream` regression that fails on pre-fix code. This PR ships none — the Slowloris fix is unpinned.

**4. No `fetchImpl` seam.** #112 threads an injectable `fetchImpl` (`:82`, used `:105`), which is what makes the regression test possible without network.

**5. Reflow direction, measured not assumed.** `fetch_helper.ts` lines over 80 columns: `main` 5, #112 6, this PR **0** (137 / 150 / 158 lines respectively). #112 sits at the width `main` already uses; this PR reformats a security-relevant file to a narrower width the repo does not use, which is most of its +69/-44.

No lockfile churn here — noting that, since it is a recurring problem on this cluster and this PR is clean on that count.

Nothing to cherry-pick onto #112. Premise re-verified unlanded: `main` (`fbd912a1`) still has `abort(new Error("fetch_helper timeout"))` at `fetch_helper.ts:87` and `clearTimeout(timer)` at `:94`, before `getReader()` at `:106`.

PR #179, opened 2026-08-16
⚡ Bolt: Replace O(N log N) boundary sorting with O(N) sequential array push

rfv-code:
💡 **What:** Replaced the `Set<number>` used to track SQL semicolon boundary indices in `src/connectors/db/lib/policy.ts` with a simple array `number[]` and sequential `push()`. This eliminated the need to convert `Set` to `Array` and perform an `O(N log N)` `.sort()` in `_splitStatements`.

🎯 **Why:** The string parsing loop in `_boundarySemicolons` only increments `i`, so any indices found are already unique and naturally monotonically increasing (sorted). Storing them in a `Set` and subsequently sorting the array was an unnecessary performance overhead, especially for larger compound queries.

📊 **Impact:** Reduces time complexity for extracting boundary constraints from `O(N log N)` back to `O(N)` since array pushes and subsequent sequential scans directly operate on natural sorted arrays without intermediate conversions.

🔬 **Measurement:** Verified via `pnpm test tests/connectors/db/policy.test.ts`. Boundary functionality operates identically, passing all existing tests.

---
*PR created automatically by Jules for task [17235269230002106715](https://jules.google.com/task/17235269230002106715) started by @rfv-code*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -10,3 +10,6 @@
 ## 2024-06-25 - Avoid array sort for Top-K in hot paths
 **Learning:** V8 engine sorting (`[...arr].sort()`) is surprisingly slow and blocks the event loop for thousands of records when fetching Top-K elements (e.g. usage statistics). It scales at O(N log N) when O(N) is sufficient for a fixed K.
 **Action:** When gathering top elements from a large data array in this codebase, manually track the top items in a single O(N) pass rather than sorting a full copy.
+## 2024-08-16 - Replace O(N log N) Set->Array sort with O(N) Array push
+**Learning:** SQL statement boundary tracking (`_boundarySemicolons`) was using a `Set<number>` to collect indices, requiring a conversion to an array and an `O(N log N)` sort afterwards in `_splitStatements`. Because the index `i` only ever increments during the scan, the indices are naturally unique and monotonically increasing.
+**Action:** Use an array and sequentially `push()` indices. This removes insertion overhead from `Set` and eliminates the need for `.sort()` downstream. When iterating loops where the index only moves forward, prefer a pre-allocated array (or sequential push) over `Set` + sort.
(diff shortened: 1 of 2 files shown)


rfv-code:
Closing as duplicate of #111, which is the standing canonical for the SQL-boundary Set/sort cluster in `policy.ts`.

The core transform is identical. Verified at both tips (`faaf9b9e` vs `22bab92f`): `Set<number>` → `number[]` (0 `Set<number>` left in either), `.add` → `.push` on the single remaining site, `.size` → `.length`, and **both** `Array.from(...).sort((a, b) => a - b)` calls dropped in favour of a direct `b1[i]` / `b2[i]` walk. Zero residual sorts in either PR — the only `.sort(` string left in this PR is inside its own explanatory comment.

The entire remaining delta is presentation. `git diff -w --ignore-blank-lines` between the two tips is 147 lines, and every one of them is a whole-file prettier reflow at a narrower print width, plus comment wording.

Formatting, measured rather than assumed — over-80-column lines in `policy.ts`:

- `main`: 31 over-80 / 1006 lines
- #111: 33 over-80 / 1006 lines
- #179: 9 over-80 / 1074 lines

So #111 sits at `main`'s prevailing width, while this PR reformats a security-relevant SQL parser to a width the repo does not otherwise use. That is the deciding factor: the reflow makes the security-relevant hunks harder to review for no behavioural gain, and #111 is a minimal +9/-9.

**One genuine merit, recorded rather than dismissed:** this PR comments *both* sites — the `boundaries` declaration inside `_boundarySemicolons` and the comparison loop in `_splitStatements` — where #111 comments only the loop. That is presentational: #111's single comment already pins the invariant that makes dropping the sort safe (the sequential scan pushes indices in monotonically increasing order, so the arrays are already sorted). Not worth a push that would risk clobbering #111's tip.

If you want the second comment, please cherry-pick just that hunk onto #111.

PR #180, opened 2026-08-17
⚡ Bolt: Replace Set with Array to eliminate O(N log N) sorting in SQL parser

rfv-code:
💡 What: Replaced a `Set<number>` with a sequentially populated `number[]` array in the `_boundarySemicolons` function, which eliminated the need for a redundant `Array.from(b).sort((a,b) => a-b)` step in `_splitStatements`.

🎯 Why: During string/SQL statement splitting, character indices are naturally processed in a sequentially increasing order (0, 1, 2...). Placing these indices into a `Set` and subsequently sorting them is redundant and introduces an unnecessary O(N log N) performance bottleneck (as well as `Set` object creation overhead) when an O(N) array push is perfectly sufficient.

📊 Impact: Reduces memory allocation (avoiding Set -> Array conversions) and eliminates an O(N log N) sorting step for every compound SQL query processed by the engine.

🔬 Measurement: Verify by executing the test suite (`pnpm test tests/connectors/db/policy.test.ts`), which ensures no regressions in the SQL statement splitting bounds logic.

---
*PR created automatically by Jules for task [5091237778420047036](https://jules.google.com/task/5091237778420047036) started by @rfv-code*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -10,3 +10,6 @@
 ## 2024-06-25 - Avoid array sort for Top-K in hot paths
 **Learning:** V8 engine sorting (`[...arr].sort()`) is surprisingly slow and blocks the event loop for thousands of records when fetching Top-K elements (e.g. usage statistics). It scales at O(N log N) when O(N) is sufficient for a fixed K.
 **Action:** When gathering top elements from a large data array in this codebase, manually track the top items in a single O(N) pass rather than sorting a full copy.
+## 2026-05-25 - Avoid Sets for sequential iteration tracking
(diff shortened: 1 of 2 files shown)


rfv-code:
Closing as duplicate of #111.

Both branches make the identical change to `src/connectors/db/lib/policy.ts`: `_boundarySemicolons` returns `number[]` instead of `Set<number>`, `.size` → `.length`, and the redundant `Array.from(b).sort((a,b) => a-b)` pair in `_splitStatements` is dropped. The source diffs are equivalent line for line.

#111 is the canonical because it is already reviewed and verified behavior-preserving (approve, ready to merge) and its diff is confined to the source file, with no `.jules/bolt.md` churn. This branch is newer, but recency is only a tiebreaker and the completed verification on #111 is the stronger signal.

The one thing unique here is the `.jules/bolt.md` learning entry ("Avoid Sets for sequential iteration tracking"). If that note is worth keeping, cherry-pick just that hunk onto #111.

Merge note carried over from the #111 review: land #94 first, then rebase, to avoid a conflict in `_splitStatements`.

PR #181, opened 2026-08-17
🛡️ Sentinel: [CRITICAL] Fix Sensitive Data Leak in Error Messages

rfv-code:
🚨 **Severity:** CRITICAL
💡 **Vulnerability:** Unscrubbed error messages (including Zod parser payloads and default generalized connector faults) were directly serialized and printed to standard outputs/errors and JSON envelopes, resulting in the potential exposure of plaintext credentials or application internals.
🎯 **Impact:** If a connector or SDK fails during operation handling sensitive arguments or failing due to misconfigured secrets, these secrets could have been leaked downstream into operational logs or via IPC API outputs.
🔧 **Fix:** Applied the `scrubSecrets()` sanitizer natively across `src/toolkit/connector.ts` within the `defaultErrorMap`, `errorEnvelope`, `writeArgErrorEnvelope`, and custom override paths in `mapAndBuildError` to safeguard all error outputs. 
✅ **Verification:** Re-ran all tests including `tests/toolkit/connector.test.ts`. Confirmed via PR testing logic that previously raw outputs are gracefully scrubbed.

---
*PR created automatically by Jules for task [6354486732164466918](https://jules.google.com/task/6354486732164466918) started by @rfv-code*


rfv-code:
Closing as duplicate of #154.

Same file (`src/toolkit/connector.ts`), same vulnerability (credential material surviving into error envelopes and stderr), same author. #154 is the canonical: it carries prior fix-pushes from earlier review rounds, its scrub sits at the `mapAndBuildError` choke point so custom `mapError` overrides are covered too, and its diff is confined to the source file — this branch also churns 4,700 lines of `pnpm-lock.yaml` with no dependency change to justify it.

**This branch did cover one thing #154 did not**, and it has now been ported rather than dropped. Scrubbing inside `defaultErrorMap` closes the zod-validation path in `run()`, which builds its `ErrorEnvelope` inline and returns before reaching `mapAndBuildError` — so a credential echoed back by a zod issue message escaped unredacted in both `message` and the hardship recorder `context`. Landed on #154 as `b6e9cbf`, with four tests; the zod case fails without the change.

One deliberate difference from this branch's version: the redaction runs **after** the heuristic classification, not before. Scrubbing first lets a secret whose plaintext contains `404` or `timeout` steer the `error_code`, which is a (minor) attacker-influenced behavior this branch introduced. There is a test for that ordering on #154.

Nothing else here is unique — the remaining diff is prettier reflow of untouched declarations.

PR #182, opened 2026-08-18
⚡ Bolt: Eliminate O(N log N) overhead in SQL boundary detection

rfv-code:
💡 What: Replaced `Set<number>` with a standard `number[]` for tracking semicolon boundaries in `_boundarySemicolons`, and removed the subsequent `Array.from(...).sort(...)` call.
🎯 Why: The sequential loop over the SQL string naturally produces monotonically increasing indices. Storing these in a Set only to immediately convert and sort them introduces unnecessary O(N log N) overhead for what is inherently an O(N) operation.
📊 Impact: Eliminates redundant sorting and Set-to-Array conversions in the hot path of the SQL policy parsing engine, improving the performance of `_splitStatements`.
🔬 Measurement: Verified by running `pnpm test tests/connectors/db/policy.test.ts`. All 59 tests continue to pass with the optimization in place.

---
*PR created automatically by Jules for task [9816159023107403100](https://jules.google.com/task/9816159023107403100) started by @rfv-code*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -10,3 +10,6 @@
 ## 2024-06-25 - Avoid array sort for Top-K in hot paths
 **Learning:** V8 engine sorting (`[...arr].sort()`) is surprisingly slow and blocks the event loop for thousands of records when fetching Top-K elements (e.g. usage statistics). It scales at O(N log N) when O(N) is sufficient for a fixed K.
 **Action:** When gathering top elements from a large data array in this codebase, manually track the top items in a single O(N) pass rather than sorting a full copy.
+## 2024-05-18 - Avoid Set for sequential index tracking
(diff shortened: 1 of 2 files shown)


rfv-code:
Closing as duplicate of #111.

Verified: the core transform is identical — `_boundarySemicolons` returns
`number[]` instead of `Set<number>`, `.add` → `.push`, `.size` → `.length`,
and both `Array.from().sort()` calls drop out of `_splitStatements`.

`git diff -w --ignore-blank-lines pr/111 pr/182 -- src/connectors/db/lib/policy.ts`
reduces to variable naming (`b1Array`/`b2Array` here vs `b1`/`b2` on #111) and
where the rationale comment sits. No behavioral delta.

#111 is kept because it places the monotonic-order rationale at the comparison
loop, which is the code that actually depends on the invariant, and it is the
branch with the trial-merge verification against #94 already on record.

For the record, the `Array.from(`/`.sort(` that a residual-marker grep reports
on this branch is the prose inside your own comment at `policy.ts:256`, not
code — this branch has zero remaining sorts.

PR #183, opened 2026-08-18
🛡️ Sentinel: [CRITICAL] Fix DoS vulnerability via Slowloris stream timeout & memory exhaustion

rfv-code:
🚨 **Severity:** CRITICAL
💡 **Vulnerability:** Unbounded HTTP fetch responses without streaming size limits (Memory Exhaustion) and premature clearing of `AbortController` timeouts before body streams are complete (Slowloris). 
🎯 **Impact:** Malicious servers could crash the application by serving massive payloads (OOM DoS) or infinitely trickling response bytes to tie up fetch threads and memory without triggering timeouts.
🔧 **Fix:** Refactored `fetchWithCaps` in `src/toolkit/fetch_helper.ts` to ensure the `clearTimeout` executes in a `finally` block wrapping the stream iteration. Migrated `fetch_attachment.ts` from naive `.arrayBuffer()` accumulation to the capped `fetchWithCaps` utility.
✅ **Verification:** Verified fixes locally via `pnpm test tests/toolkit/fetch_attachment.test.ts tests/toolkit/fetch_helper.test.ts`, both of which pass completely with 13/13 unit tests. Formatted correctly via Prettier. Temporary ad-hoc files cleaned up.

---
*PR created automatically by Jules for task [2704946545151894730](https://jules.google.com/task/2704946545151894730) started by @rfv-code*


rfv-code:
Closing as duplicate of #112.

Verified: the same fix — `fetch_attachment.ts` routed through `fetchWithCaps`
so the byte cap applies mid-stream, and the `fetch_helper.ts` read loop moved
inside the timed `try` so a Slowloris stall cannot outlive the timeout.

#112 is kept for two reasons this branch does not match:

1. **Tests.** #112 ships 85 lines of regression coverage in
   `tests/toolkit/fetch_helper.test.ts`, including a stalled `ReadableStream`
   that hangs on pre-fix code. This branch ships none.
2. **The abort contract.** #112 carries `c04ec6f`, which makes the timeout
   abort with `DOMException(..., "AbortError")` rather than a plain `Error`,
   matching what the docs promise. This branch still has main's behavior
   (`AbortError` ×2 / `DOMException` ×1, unchanged from `fbd912a1`).

The one genuine difference is that this branch threads `fetchImpl` as an
options-object field where #112 uses a 4th positional parameter. That form is
arguably cleaner, and it has been re-proposed on several regenerations of this
fix, but it is a style change rather than a behavioral one and adopting it now
would churn #112's call sites and tests for no functional gain.

PR #184, opened 2026-08-19
⚡ Bolt: Replace Set with Array for sequential boundary tracking

rfv-code:
💡 **What:** Replaced the use of `Set<number>` with `number[]` in `_boundarySemicolons` and eliminated the subsequent `Array.from(...).sort(...)` calls in `_splitStatements`.

🎯 **Why:** The parsing loop in `_boundarySemicolons` processes the SQL string strictly sequentially from left to right. Any semicolon boundary discovered is inherently greater than the last. Accumulating these in a Set and then sorting them is redundant and introduces an unnecessary `O(N log N)` computational overhead to a critical, frequently-called parsing path. 

📊 **Impact:** Reduces memory allocation overhead (Arrays vs Sets) and changes the statement-splitting time complexity for boundary sorting from `O(N log N)` to strict `O(N)`.

🔬 **Measurement:** Code logic confirmed by existing test suite. `pnpm test tests/connectors/db/policy.test.ts` passes successfully, validating that statement boundaries are still parsed accurately across all dialects.

---
*PR created automatically by Jules for task [1706588776754965364](https://jules.google.com/task/1706588776754965364) started by @rfv-code*

### DIFF
--- a/src/connectors/db/lib/policy.ts
+++ b/src/connectors/db/lib/policy.ts
@@ -251,8 +251,11 @@ function _boundarySemicolons(
   sql: string,
   treatBackslashAsEscape: boolean,
   dialect: SqlDialect = "generic",
-): Set<number> {
-  const boundaries = new Set<number>();
+): number[] {
+  // ⚡ Bolt: Replace Set<number> with number[] for sequential loop parsing.
+  // Pushing sequentially inherently keeps the array sorted and avoids an
+  // unnecessary O(N log N) sorting step later in _splitStatements.
+  const boundaries: number[] = [];
   let inString: string | null = null; // Either null, "'", '"', or '`'
 
   for (let i = 0; i < sql.length; i++) {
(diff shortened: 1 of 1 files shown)


rfv-code:
Closing as duplicate of #111.

`git diff pr/111 pr/184 -- src/connectors/db/lib/policy.ts` is the whole delta, and it is presentation only:

- **Naming.** #184 names the locals `b1Array`/`b2Array` at the declarations; #111 names them `b1`/`b2` and binds `const boundaries = b1` before the split loop.
- **Comment placement.** #184 puts the rationale on `_boundarySemicolons`; #111 puts it on the comparison loop in `_splitStatements`, where it pins the monotonic-push invariant that makes the index-wise compare safe.

The transform itself is identical in both: `Set<number>` → `number[]`, `.add` → `.push`, `.size` → `.length`, and both `Array.from(...).sort(...)` calls dropped. #111 is the standing canonical for this cluster and has carried a long-form review since 2026-07-14.

One thing #184 does better and #111 does not: it comments the **declaration** site, so a reader who lands in `_boundarySemicolons` first sees why the return type is an array. If you would rather keep that, cherry-pick the three comment lines onto #111 — it conflicts with nothing.

For the record, #184 is also the cleanest shape this cluster has produced: +9/-8, no whole-file reflow, no `.jules/*` churn, and no `package-lock.json` hunk.

PR #189, opened 2026-08-22
⚡ Bolt: Optimize SQL statement parsing performance

rfv-code:
💡 **What:** Refactored `_boundarySemicolons` in `src/connectors/db/lib/policy.ts` to directly return a `number[]` instead of a `Set<number>`. Removed the subsequent `Array.from(set).sort((a,b) => a-b)` overhead in `_splitStatements`.
🎯 **Why:** The parsing loop iterates linearly over the `sql` string. Therefore, the indices of semicolons discovered are inherently strictly monotonically increasing. Pushing them to an array naturally produces a sorted list without requiring explicit set deduplication or sorting. The redundant O(N log N) sorting was blocking the event loop pointlessly.
📊 **Impact:** Eliminates O(N log N) sorting overhead and reduces memory allocation (Set vs Array) during SQL statement parsing in the critical path of DB policy evaluation.
🔬 **Measurement:** Code was verified via existing unit tests (`pnpm test tests/connectors/db/policy.test.ts`), which continue to pass flawlessly. Metrics: O(N log N) -> O(N) traversal.

---
*PR created automatically by Jules for task [425038954846885966](https://jules.google.com/task/425038954846885966) started by @rfv-code*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -10,3 +10,6 @@
 ## 2024-06-25 - Avoid array sort for Top-K in hot paths
 **Learning:** V8 engine sorting (`[...arr].sort()`) is surprisingly slow and blocks the event loop for thousands of records when fetching Top-K elements (e.g. usage statistics). It scales at O(N log N) when O(N) is sufficient for a fixed K.
 **Action:** When gathering top elements from a large data array in this codebase, manually track the top items in a single O(N) pass rather than sorting a full copy.
+## 2024-05-18 - Redundant Sorting Overhead in Sequential Parsing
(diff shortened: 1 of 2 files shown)


rfv-code:
Closing as duplicate of #111.

Byte-identical change to #111's `policy.ts` hunk (`_boundarySemicolons` returns `number[]`, `_splitStatements` drops `Array.from().sort()`) — verified with `git diff` against both tips. #111 is kept as canonical because it carries the same optimization plus an inline comment recording *why* the sort is safe to drop (the boundaries are already populated in monotonic order), and it has a prior structured review.

If this branch covers anything #111 does not, please cherry-pick onto #111.

PR #192, opened 2026-08-23
🛡️ Sentinel: [HIGH] Fix Slowloris and memory exhaustion vulnerabilities in fetch primitives

rfv-code:
🚨 **Severity**: HIGH

💡 **Vulnerability**: 
1. **Slowloris DoS**: In `fetchWithCaps`, the `AbortController` timer for the fetch timeout was being prematurely cleared inside a `finally` block immediately after the initial HTTP headers were received. This left the subsequent stream reading loop (`reader.read()`) unbounded by time, meaning a malicious server could accept the connection and then trickle a response body indefinitely (Slowloris attack) tying up connector resources.
2. **Memory Exhaustion**: In `fetchAttachment`, the code managed the timeout manually and read the HTTP response directly into memory using `await response.arrayBuffer()`, bypassing the streaming caps implemented in `fetchWithCaps`. If the server lied about the `content-length`, it could trigger an Out-of-Memory (OOM) error by returning a massive file.

🎯 **Impact**: An attacker or a malfunctioning server could exhaust application memory or hang HTTP connector requests indefinitely, leading to degraded performance or complete denial of service.

🔧 **Fix**: 
- Moved the `clearTimeout()` in `fetchWithCaps` to execute only *after* the stream has been fully read.
- Updated `FetchCapsOptions` to accept a custom `fetchImpl` to cleanly pass dependency-injected fetch functions without temporary global state overrides.
- Refactored `fetchAttachment` to use `fetchWithCaps`, taking advantage of its streaming bounds and the newly secured timeout logic, and removed the redundant local `AbortController` and `response.arrayBuffer()` logic.

✅ **Verification**: 
- Simulated Slowloris test script proved the HTTP response times out appropriately even if it hangs halfway through a streamed response.
- `pnpm test` confirms that `fetch_helper` and `fetch_attachment` tests pass cleanly without regression.

---
*PR created automatically by Jules for task [8471232827733238899](https://jules.google.com/task/8471232827733238899) started by @rfv-code*


rfv-code:
Closing as duplicate of #112.

Both branches make the same core fix: move the body-streaming loop inside the timed `try` in `fetchWithCaps`, so `clearTimeout` only fires in the `finally` after the body drains. That closes the Slowloris window where a server sends headers promptly then drips the body. Both also add a `fetchImpl` seam so `fetch_attachment.ts` can route through `fetchWithCaps` (this branch passes it as `caps.fetchImpl`, #112 as a fourth positional argument — equivalent).

#112 is the canonical for three reasons:

1. **It carries a behavior fix this branch does not.** #112 aborts with `new DOMException("fetch_helper timeout", "AbortError")`; this branch keeps `new Error("fetch_helper timeout")`. Both docblocks advertise `AbortError`, and callers do the conventional `err.name === "AbortError"` check. With a plain `Error`, a timeout is misclassified as a generic network failure — and now that the timer stays armed through the body stream, that misclassification reaches a second code path.
2. **It has a regression test.** #112 adds `tests/toolkit/fetch_helper.test.ts`; this branch adds no test for the new streaming-timeout behavior.
3. **This branch adds a whole-file Prettier reformat** on top of the fix, which enlarges the diff on two security-relevant files without changing behavior.

If you believe this branch covers ground #112 does not, please cherry-pick onto #112.

PR #198, opened 2026-08-27
🛡️ Sentinel: [HIGH] Fix Slowloris DoS vulnerability in fetch clients

rfv-code:
🚨 **Severity:** HIGH
💡 **Vulnerability:** Slowloris DoS vulnerability in `fetch_helper` and `fetch_attachment`. The HTTP timeout timer was being cleared immediately after the response headers resolved, leaving the subsequent body streaming (`response.arrayBuffer()` or `reader.read()`) unbounded. A malicious or degraded server could trickle the response body indefinitely (e.g., 1 byte per minute) and hang the connector process.
🎯 **Impact:** Potential exhaustion of connector concurrency slots leading to denial of service, as processes wait indefinitely for data that never finishes arriving.
🔧 **Fix:** Expanded the `try...finally { clearTimeout(timer); }` block to wrap the *entire* request lifecycle, including the consumption of the response body streams.
✅ **Verification:** Ad-hoc Slowloris test scripts confirmed the 2-second timeout successfully interrupts a stalled body read, and the `vitest` suite (`pnpm test`) continues to pass successfully without regressions.

---
*PR created automatically by Jules for task [16063830792240023210](https://jules.google.com/task/16063830792240023210) started by @rfv-code*


rfv-code:
Closing as duplicate of #112.

Both PRs make the same fix — keep the timeout timer armed across the body read so `clearTimeout` only fires in the `finally` after the stream drains. #112 is the canonical because it is a strict superset:

- Same Slowloris fix in `fetchWithCaps` (`src/toolkit/fetch_helper.ts`, commit f6718d3 "keep timeout armed during body stream to close Slowloris window").
- Also routes `fetchAttachment` through `fetchWithCaps`, so the size cap runs on the *stream* instead of trusting `content-length` and then buffering the full body. This PR keeps `fetchAttachment` on its own `doFetch` path, which still buffers an unbounded body from a server that lies about `content-length`.
- Also fixes the abort reason to a `DOMException` named `AbortError` (commit c04ec6f), so callers doing the conventional `err.name === "AbortError"` check no longer misclassify a timeout as a network failure.
- Has 85 lines of new tests in `tests/toolkit/fetch_helper.test.ts`. This PR has none.

The remainder of this diff is a whole-file Prettier reformat of `fetch_helper.ts` and `fetch_attachment.ts` plus `package-lock.json` churn, with no behavior change.

Nothing here is exclusive to this PR, so there is nothing to cherry-pick onto #112.

PR #200, opened 2026-08-28
🛡️ Sentinel: [CRITICAL] Fix Slowloris DoS in fetchWithCaps

rfv-code:
🚨 Severity: CRITICAL
💡 Vulnerability: The `fetchWithCaps` timeout was cleared right after headers resolved. Malicious servers could bypass the timeout by delaying the response body stream infinitely (Slowloris DoS).
🎯 Impact: Prevents indefinite hangs and memory leak vulnerabilities during outbound network requests fetching external assets.
🔧 Fix: Wrapped the body reading logic in a `try...finally` block and moved `clearTimeout` to the `finally` block so the timeout covers the complete request execution lifetime. Added testing for slow streams.
✅ Verification: Ran `pnpm test` and ensured `fetch_helper.test.ts` passes the Slowloris test behavior.

---
*PR created automatically by Jules for task [16144689661850014635](https://jules.google.com/task/16144689661850014635) started by @rfv-code*


rfv-code:
Closing as duplicate of #112.

Same Slowloris fix — keep the timeout armed through the body stream by holding the streaming loop inside the timed `try` and clearing only in the `finally`. This branch is 1 commit off the merge base; #112 has 3 commits and a strictly larger scope.

What #112 has that this branch does not:
- `fetchImpl: typeof fetch = globalThis.fetch` — the injection seam the timeout tests need.
- Timeouts abort with `new DOMException("fetch_helper timeout", "AbortError")` instead of a plain `Error` (`c04ec6f`). With a plain `Error`, the conventional `err.name === "AbortError"` check misclassifies a timeout as a generic network failure — and now that the timer stays armed through the body, that misclassification reaches a second path.
- `fetch_attachment.ts` routed through `fetchWithCaps`, so the mid-stream byte cap applies there too. This branch does not touch that file.
- 4 new test cases against this branch's 1: header-phase `AbortError`, body-stream `AbortError`, external abort reason propagated unchanged, plus the Slowloris case.

The remaining difference in `fetch_helper.ts` is a whole-file Prettier reformat.

If a hunk here is missing from #112, please cherry-pick it onto #112.

PR #201, opened 2026-08-29
🛡️ Sentinel: [CRITICAL] Fix fetch Slowloris & memory DoS

rfv-code:
🚨 Severity: CRITICAL
💡 Vulnerability: 
1. The `fetchWithCaps` helper cleared its timeout immediately after the HTTP headers were fetched, meaning the response body could trickle in indefinitely, enabling a Slowloris-style denial-of-service. 
2. The `fetch_attachment` utility bypassed `fetchWithCaps` entirely and called raw `fetch` followed by an unconstrained `response.arrayBuffer()`, exposing the connector process to memory exhaustion DoS if a server sent a massive body despite a small advertised `content-length`.
🎯 Impact: Both issues could allow an attacker controlling an external server (e.g. Jira/Confluence/Notion attachment fetch) to stall or crash the agent process via memory exhaustion or stalled open connections.
🔧 Fix: 
- Wrapped the entire streaming process in `fetchWithCaps` with a `finally` block to clear the timer only after stream consumption completes.
- Refactored `fetch_attachment` to rely on `fetchWithCaps` for size/timeout limits.
✅ Verification: `pnpm test tests/toolkit` succeeds and verifies mock integrations work correctly.

---
*PR created automatically by Jules for task [8328617957930540160](https://jules.google.com/task/8328617957930540160) started by @rfv-code*


rfv-code:
Closing as duplicate of #112.

Same fix, and a strict subset of it. Both route `fetchAttachment` through `fetchWithCaps` and move the `clearTimeout` into a `finally` that runs only after the body stream is drained, closing the Slowloris window where a server sends headers promptly and then drips the body.

#112 is the canonical — three commits, plus 85 lines of tests in `tests/toolkit/fetch_helper.test.ts`, plus one fix this branch does not have:

```
timeoutCtl.abort(new DOMException("fetch_helper timeout", "AbortError"))
```

This branch keeps `new Error("fetch_helper timeout")`. The abort reason propagates verbatim to whatever `fetch`/`reader.read()` rejects with, and the docblocks advertise `AbortError`, so a caller doing the conventional `err.name === "AbortError"` check misclassifies a timeout as a generic network failure. Keeping the timer armed through the body read — which both PRs do — puts that misclassification on a second path, so it matters more after this change, not less.

The only other difference is where `fetchImpl` is threaded (`FetchCapsOptions.fetchImpl` here, a fourth positional parameter on #112). That is a style choice with no behavioral difference, and it is not worth losing #112's tests over.

No exclusive scope found here relative to #112.

PR #202, opened 2026-08-29

rfv-code:
💡 **What:** 
Replaced `Set` usage with `Array.push()` in `_boundarySemicolons` and removed the subsequent `Array.from().sort()` conversions in `_splitStatements`.

🎯 **Why:** 
When iterating character-by-character over a SQL string to find semicolon boundaries, the collected indices are naturally unique and monotonically increasing. Using a `Set` and subsequently sorting the keys is unnecessary and incurs O(N log N) overhead. Pushing directly to an array achieves the exact same result in strict O(N) time with significantly less object allocation.

📊 **Impact:** 
Removes O(N log N) sorting overhead and reduces garbage collection pressure by avoiding intermediate `Set` and `Array.from` object allocations during SQL policy parsing, which occurs frequently on the hot path for all incoming DB queries.

🔬 **Measurement:** 
Unit tests for the policy engine (`tests/connectors/db/policy.test.ts`) continue to pass successfully, confirming that statement boundary behaviors remain identical.

---
*PR created automatically by Jules for task [11949730837333157106](https://jules.google.com/task/11949730837333157106) started by @rfv-code*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -10,3 +10,6 @@
 ## 2024-06-25 - Avoid array sort for Top-K in hot paths
 **Learning:** V8 engine sorting (`[...arr].sort()`) is surprisingly slow and blocks the event loop for thousands of records when fetching Top-K elements (e.g. usage statistics). It scales at O(N log N) when O(N) is sufficient for a fixed K.
 **Action:** When gathering top elements from a large data array in this codebase, manually track the top items in a single O(N) pass rather than sorting a full copy.
+## 2024-07-28 - Sequential string processing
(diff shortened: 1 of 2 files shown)


rfv-code:
Closing as duplicate of #111.

Identical change to `_boundarySemicolons` / `_splitStatements`: `Set<number>` → `number[]`, `.add` → `.push`, and dropping the two `Array.from(...).sort()` calls plus the `.size` comparison in favour of `.length`. #111 is the canonical — same optimization, two commits, and two tests in `tests/connectors/db/policy_security.test.ts`:

- one pinning the boundary parity check at the *position* level, not just the count (`SELECT 'a\\';x'; SELECT 2` — both escape modes find one semicolon, at different offsets; a length-only comparison would split at the attacker-chosen offset),
- one walking a four-boundary script to pin that the ordering assumption this optimization relies on actually holds.

Those tests are what make the `sort()` removal safe to land, and this branch has none. The remainder of its 111 insertions is a whole-file Prettier reformat.

No exclusive scope found here relative to #111. If I missed something, please cherry-pick it onto #111.

PR #205, opened 2026-08-31

rfv-code:
💡 **What**: Replaced the `Set` with an `Array` (`number[]`) for tracking SQL statement boundary indices in `_boundarySemicolons` within `src/connectors/db/lib/policy.ts`. Removed the subsequent `Array.from(...).sort(...)` calls since boundaries appended sequentially are inherently sorted.
🎯 **Why**: The original implementation used a `Set` to collect semicolon boundary indices, and then performed an unnecessary V8 engine array conversion and sorting `[...set].sort()` at the end. Since the loop always increments sequentially, collecting indices into a simple array inherently produces unique, monotonically increasing values, avoiding the O(N log N) sorting overhead.
📊 **Impact**: Eliminates O(N log N) array sorting overhead during SQL statement boundary parsing, replacing it with an O(N) array append strategy, improving performance on long, compound SQL strings.
🔬 **Measurement**: Verify that all 59 tests in `tests/connectors/db/policy.test.ts` pass, ensuring that statement boundaries are identical to the previous implementation.

---
*PR created automatically by Jules for task [550609323464456390](https://jules.google.com/task/550609323464456390) started by @rfv-code*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -10,3 +10,6 @@
 ## 2024-06-25 - Avoid array sort for Top-K in hot paths
 **Learning:** V8 engine sorting (`[...arr].sort()`) is surprisingly slow and blocks the event loop for thousands of records when fetching Top-K elements (e.g. usage statistics). It scales at O(N log N) when O(N) is sufficient for a fixed K.
 **Action:** When gathering top elements from a large data array in this codebase, manually track the top items in a single O(N) pass rather than sorting a full copy.
+## 2026-08-31 - Array over Set for Sequential Tracking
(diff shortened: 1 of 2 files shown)


rfv-code:
Closing as duplicate of #111.

Both PRs make the identical behavioral change to `src/connectors/db/lib/policy.ts`: `_boundarySemicolons` returns `number[]` instead of `Set<number>`, and `_splitStatements` compares the arrays directly instead of `Array.from(set).sort()`. #111 is chosen as canonical because it adds 23 lines of coverage in `tests/connectors/db/policy_security.test.ts` and already carries fix-pushes from prior review rounds.

This branch is 1 commit with no tests, and most of its +116/-46 is a whole-file Prettier reformat rather than the optimization.

If you think a case here is missed by #111, please cherry-pick that hunk onto `bolt/optimize-sql-boundary-parsing-3654430659625864006`.

PR #206, opened 2026-09-01
🛡️ Sentinel: [HIGH] Fix Slowloris DoS vulnerability in fetch_helper

rfv-code:
🚨 **Severity**: HIGH

💡 **Vulnerability**: A Slowloris Denial of Service (DoS) vulnerability existed in `fetchWithCaps` where the internal `AbortController` timeout was cleared immediately after the initial HTTP fetch resolved (headers received). This allowed a malicious or compromised endpoint to trickle the response body over an indefinite amount of time, keeping the memory-stream reading process open permanently and exhausting resources without triggering the timeout.

🎯 **Impact**: An attacker or rogue endpoint could cause prolonged resource and memory exhaustion for the agent/connector by intentionally delaying the delivery of response stream chunks, potentially leading to service degradation or a complete crash.

🔧 **Fix**: Moved the `clearTimeout(timer)` call from immediately after the initial fetch into a `finally` block encompassing the entire function. The timeout now safely encapsulates the full lifecycle of the request, including the subsequent `reader.read()` loop streaming the response body.

✅ **Verification**: Added a new test unit in `tests/toolkit/fetch_helper.test.ts` that mocks a slow-streaming stream that trickles byte chunks. Verified the test appropriately `rejects.toThrow()` due to timeout limits correctly applying during stream reading. Evaluated with `pnpm test` and `pnpm typecheck` locally to confirm code compliance.

---
*PR created automatically by Jules for task [12759960658627032934](https://jules.google.com/task/12759960658627032934) started by @rfv-code*


rfv-code:
Closing as duplicate of #112.

Both PRs make the same core fix in `src/toolkit/fetch_helper.ts`: move the streaming body read inside the `try` so `clearTimeout` only fires in the `finally` after the body is drained, closing the Slowloris window where a server sends headers promptly then drips the body.

#112 is a strict superset: same fix, plus it aborts with `new DOMException(..., "AbortError")` instead of a plain `Error` (callers do the conventional `err.name === "AbortError"` check, and the now-longer-armed timer makes that misclassification reachable on a second path), plus it consolidates `fetch_attachment.ts` onto `fetchWithCaps` so the cap logic exists once. It also carries 85 lines of new tests against this PR's 28.

Nothing here to cherry-pick onto #112.

PR #211, opened 2026-09-03
🛡️ Sentinel: [CRITICAL] Fix Slowloris DoS vulnerability in fetch streams

rfv-code:
🚨 Severity: CRITICAL
💡 Vulnerability: Slowloris / Slow-read Denial of Service (DoS) in streaming HTTP fetches where a malicious server can trickle a response body indefinitely because the timeout timer is cleared as soon as headers are received.
🎯 Impact: An attacker could exhaust server resources by tying up memory and keeping promises hanging indefinitely, leading to resource exhaustion or system crash.
🔧 Fix: Moved `clearTimeout(timer)` to encompass the entire body stream processing loops (`reader.read()` and `response.arrayBuffer()`) within the `try...finally` block in `fetch_helper.ts` and `fetch_attachment.ts`.
✅ Verification: Ran `pnpm typecheck` and the full `vitest` suite with `pnpm test` successfully. Prettier formatting has been applied.

Also added a journal entry regarding proper placement of `clearTimeout` for streaming APIs.

---
*PR created automatically by Jules for task [11869514267982242961](https://jules.google.com/task/11869514267982242961) started by @rfv-code*


rfv-code:
Closing as duplicate of #112.

Both close the same Slowloris window by moving the body-read loop inside the `try` so `clearTimeout` only fires in the `finally` after the body drains. I diffed both against `origin/main`; #112 is a strict superset:

- It routes `fetchAttachment` through `fetchWithCaps` instead of leaving the duplicated cap logic in place, so the streaming cap covers attachments too. This branch reorders `fetchAttachment`'s own try/finally but leaves it on `response.arrayBuffer()`, which does not enforce the incremental byte cap.
- It threads a `fetchImpl` parameter through `fetchWithCaps`, which is what makes the path testable.
- It aborts with `new DOMException("fetch_helper timeout", "AbortError")` rather than a plain `Error`, so callers doing the conventional `err.name === "AbortError"` check classify a timeout correctly. This branch keeps the plain `Error`.
- It has `tests/toolkit/fetch_helper.test.ts`; this branch has no test.

This branch also bundles a whole-file Prettier reformat of both files. Nothing here is exclusive to it.

PR #216, opened 2026-09-05
⚡ Bolt: Replace Set sorting with O(N) array tracking for SQL statement parsing

rfv-code:
💡 What:
Replaced the `Set<number>` used in `_boundarySemicolons` with a `number[]` array using `.push()`. In `_splitStatements`, removed the `Array.from(set).sort()` calls.

🎯 Why:
The string parsing loop's index `i` inherently only advances forward. Therefore, indices added to the collection are naturally deduplicated and already perfectly sorted in strictly increasing order. The use of a Set, followed by array conversion and sorting, introduced unnecessary O(N log N) performance overhead for an operation that is inherently O(N).

📊 Impact:
Removes redundant O(N log N) sorting steps and the overhead of Set tracking in a hot path related to SQL policy classification, converting it to an exact O(N) operation.

🔬 Measurement:
Run the test suite `pnpm test tests/connectors/db/policy.test.ts` to ensure that statement boundaries continue to be accurately parsed. Test suite passes successfully, proving semantic parity with improved execution efficiency.

---
*PR created automatically by Jules for task [244505090937722012](https://jules.google.com/task/244505090937722012) started by @rfv-code*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -10,3 +10,6 @@
 ## 2024-06-25 - Avoid array sort for Top-K in hot paths
 **Learning:** V8 engine sorting (`[...arr].sort()`) is surprisingly slow and blocks the event loop for thousands of records when fetching Top-K elements (e.g. usage statistics). It scales at O(N log N) when O(N) is sufficient for a fixed K.
 **Action:** When gathering top elements from a large data array in this codebase, manually track the top items in a single O(N) pass rather than sorting a full copy.
+## 2024-05-18 - Sequential string loop sets vs arrays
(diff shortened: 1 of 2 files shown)


rfv-code:
Closing as duplicate of #111. The `src/connectors/db/lib/policy.ts` change here is byte-identical to #111's after normalizing comments and whitespace (same `Set<number>` → `number[]` conversion, same removal of the two `Array.from().sort()` calls, same monotonic-order justification) — only the wording of the explanatory comment differs. #111 is the canonical: it carries the same optimization plus two regression tests in `tests/connectors/db/policy_security.test.ts` that this PR does not have — one pinning the boundary-parity check at the *position* level rather than the count (`SELECT 'a\\';x'; SELECT 2` must still throw), and one pinning ordered multi-boundary splitting. Dropping those tests is exactly what makes the `sort()` removal unsafe to review. If you believe this branch covers ground #111 does not, please cherry-pick onto #111.

PR #223, opened 2026-09-06
🛡️ Sentinel: [MEDIUM] Fix sensitive data leak in error envelopes

rfv-code:
🚨 **Severity:** MEDIUM
💡 **Vulnerability:** Unhandled exceptions and validation failures mapped to an `ErrorEnvelope` were emitting the raw error string to CLI standard output/error or API responses without sanitization. This could leak credentials, API keys, or database URIs if they are embedded in the error text (e.g., from a failed network fetch or parse error).
🎯 **Impact:** Potential leakage of secrets to end-users or application logs depending on how the connector output is surfaced.
🔧 **Fix:** Applied the existing `scrubSecrets` utility (which was previously only used for audit logging) directly to the message strings constructed in `errorEnvelope`, `mapAndBuildError`, and `writeArgErrorEnvelope` within `src/toolkit/connector.ts`.
✅ **Verification:** Verified via code review and full test suite run (`pnpm test`) that changes are safe and the logic appropriately handles the sanitized strings.

---
*PR created automatically by Jules for task [7205867099780408375](https://jules.google.com/task/7205867099780408375) started by @rfv-code*


rfv-code:
Closing as duplicate of #207.

#207 is the canonical credential-redaction PR for `src/toolkit/connector.ts`: 47 commits, cumulative fix-pushes from the daily review task, and a cross-product redaction test suite. This branch is 1 commit against the same merge base.

I checked every call site this PR scrubs against #207's tip before closing, rather than trusting the file list:

| This PR scrubs | #207 |
|---|---|
| `Policy config error` | covered — `policyLoadError` is scrubbed at assignment (connector.ts:907) |
| `mapped.message` (param validation) | covered (connector.ts:992) |
| `writeArgErrorEnvelope` | covered (connector.ts:1301) |
| `errorEnvelope()` body | covered at both callers (connector.ts:1039, :1074) |
| `override.message` / `def.message` | covered (connector.ts:1492), plus a second input-echo layer |
| **`Unknown action '<action>'`** | **not covered — this PR was right** |

So this branch was not a pure duplicate: it found one real gap. I landed that scrub on #207 in 82b5dcc, with a test that fails against the pre-fix source, rather than dropping it with this PR. #207 is now a strict superset.

One note on the finding itself: `scrubSecrets` recognises shapes, so `api_key=<secret>` in the action slot is redacted but a bare token is not. That residue is stated in the commit message on #207 — this path returns before `params`/`credentials` exist, so the input-echo rule cannot reach it.

PR #226, opened 2026-09-07
🛡️ Sentinel: [HIGH] Fix Slowloris DoS vulnerability in fetch streams

rfv-code:
🚨 **Severity**: HIGH
💡 **Vulnerability**: The `fetchWithCaps` and `fetchAttachment` wrappers in the toolkit called `clearTimeout` immediately after the initial HTTP headers were resolved. This left the stream processing loop (reading the response body) completely unbounded.
🎯 **Impact**: A malicious server could exploit this by slowly trickling the response body, keeping the connection open indefinitely and causing a Slowloris-style Denial of Service (DoS) memory/connection exhaustion attack.
🔧 **Fix**: Moved the `clearTimeout` call into a `finally` block that wraps the entire fetch and stream reading execution path in both files.
✅ **Verification**: Run `pnpm test tests/toolkit/fetch_helper.test.ts tests/toolkit/fetch_attachment.test.ts` to ensure the timeout semantics and byte limits still function correctly.

---
*PR created automatically by Jules for task [17938220973586703682](https://jules.google.com/task/17938220973586703682) started by @rfv-code*


rfv-code:
Closing as duplicate of #112 — but you were right, and the fix is now landed there.

**Your finding was correct and it was missing.** `fetchWithCaps` cleared its abort timer immediately after headers resolved, so the `getReader().read()` drain loop ran with no timeout at all. A server that answers headers promptly and then drips the body holds the read open forever.

**Why it was missing is worth recording.** #112 already fixed this in `f6718d3` (2026-07-20), and a later regeneration of that same branch reverted it — along with `c04ec6f` (timeout aborts with a `DOMException` named `AbortError`, not a plain `Error`) and both commits' test cases in `tests/toolkit/fetch_helper.test.ts`. The test file dropped back to its `main` content, so the loss did not show in the PR's file list.

**Restored on #112** as `2aea09a` + `8860521`. Verified there:

- `tests/toolkit/fetch_helper.test.ts`: 9 → 13 tests, including `keeps the timeout armed during the body stream (Slowloris)`.
- Mutation check — move `clearTimeout(timer)` back to just after the fetch and 2 tests hang and fail (60 s). The guard is real, not decorative.
- Full suite `2752 passed`, `tsc` clean.

**Why #112 stays canonical rather than this branch:** it also removes the duplicated fetch/cap/decode block from `fetch_attachment.ts` (-21 lines) so attachments inherit the streaming cap instead of `response.arrayBuffer()`-ing the whole payload, and it adds a `fetchImpl` injection point that makes the timeout paths testable — neither is here. The residual difference in this branch is a whole-file Prettier reformat of two security-sensitive files.

Nothing left to cherry-pick.

PR #227, opened 2026-09-08
🛡️ Sentinel: [HIGH] Fix raw error message credential leak

rfv-code:
🚨 Severity: HIGH
💡 Vulnerability: Raw error messages (e.g., from Zod validation failures, config load exceptions, or HTTP exceptions) were directly embedded in `ErrorEnvelope`s and written to standard output/error, potentially exposing sensitive tokens, configuration keys, or database credentials within the raw messages.
🎯 Impact: Anyone with access to the CLI output, or upstream services aggregating error envelopes, could unintentionally view leaked credentials embedded in unhandled or raw validation error strings.
🔧 Fix: Used the existing `scrubSecrets` utility across critical toolkit files (like `connector.ts` and `http_client.ts`) to wrap any ingested raw `err.message` before emitting it inside Envelopes or to standard streams.
✅ Verification: Ran unit tests for the modified files (`pnpm test tests/toolkit/connector.test.ts` and `tests/toolkit/http_client.test.ts`) to ensure the changes did not impact valid operations and that error envelopes retain correct structure. Types were also verified with `pnpm run typecheck`.

---
*PR created automatically by Jules for task [2005657108738783495](https://jules.google.com/task/2005657108738783495) started by @rfv-code*


rfv-code:
Both review items are fixed in 8ba88ba; threads resolved.

**P1 — scrub custom mapper output at the connector boundary.** Confirmed and fixed as suggested. `mapAndBuildError` copied `override.message` straight into the returned envelope (`src/toolkit/connector.ts:823` -> `:859`); only the hardship recorder and the resolution-hint facts were scrubbed. Since `defaultErrorMap` already scrubs on this branch, the override path was the single remaining route to an unredacted envelope.

Fixed centrally in `mapAndBuildError` rather than per-mapper, as you suggested. Measuring the class: nine `mapError` implementations exist — the three you named (aws:315, gcp:295, db/connector:172) return `err.message` verbatim, and six go through `mapHttpError`, which this PR already scrubs inside the mapper. The central scrub covers all nine plus any future one. Re-scrubbing the six is a no-op; there is a new test pinning that idempotence so `password='[REDACTED]'` is not double-mangled.

**P2 — scrub synchronous usage-report argument errors.** Confirmed: `parseArgs` throws synchronously, `main().catch` never sees it, and the inner catch printed the message directly.

I implemented your suggestion, then measured whether it was sufficient on its own. It is not. `scrubSecrets` is shape-based — it keys on quoted `key = "value"` pairs — so it redacts your example but passes bare tokens through. Measured:

| stderr input | after `scrubSecrets` |
|---|---|
| `--format "password='secret'"` | `password='[REDACTED]'` — caught |
| `--format hunter2` | `hunter2` — **leaks** |
| `Unknown arg: --token=ghp_AAAA...` | unchanged — **leaks** |

The `--token=` case is a bigger miss than the one reported: an unquoted value never matches `SENSITIVE_SQUOTE_RE`/`SENSITIVE_DQUOTE_RE`.

So the fix applies the rule this codebase already uses for rejected caller tokens in `src/hub/cli.ts` and `src/toolkit/extract_multimodal.ts` — name the accepted values, never echo the rejected one — and keeps your scrub in the catch as defence in depth for a future message that must carry caller text.

**Verification.** 4 new tests, each confirmed to fail against the unfixed code by reverting the fix and re-running. Full suite 2806 passing (baseline 2802 on this branch), `tsc` clean.

Note the first `vitest run` in a fresh worktree reports ~55 file-level collection failures until `npx tsc` has run — the integration tests load the compiled CLI from `dist/`. Build first, then the suite is green.

rfv-code:
## Round 2 — one fixed here, two belong to #207 and are already implemented there

`014ae04`. All three threads resolved. Full suite **2807 passing**, tsc clean.

### Fixed here: classify before scrub (P2) — and it was mine

This one is a regression **this PR introduced**, not a pre-existing gap: the scrub moved above the classification. Confirmed exactly as described, and measured:

| raw message | scrubbed | from raw | from scrubbed |
|---|---|---|---|
| `Authorization: Bearer abc 401 Unauthorized` | `Authorization: Bearer [REDACTED]` | AUTH_ERROR | **UNKNOWN → CONNECTION_ERROR** |
| `authorization: Bearer abc\n401 Unauthorized` | keeps the `401` line | AUTH_ERROR | AUTH_ERROR |
| `Authorization=Bearer abc forbidden` | `Authorization=Bearer [REDACTED]` | AUTH_ERROR | **UNKNOWN → CONNECTION_ERROR** |

Worth adding to your report: CONNECTION_ERROR is in `RETRIABLE_CODES`, so the misclassification also flipped `retriable` to true — an auth failure retried as a network blip. Only the *flattened* one-line form is affected, which is why the existing tests missed it.

Fixed by classifying from the raw message and returning the scrubbed one. Test verified to fail against the swapped order.

### Not fixed here: the other two are #207's, already landed there

Both remedies you ask for exist on `fix/connector-credential-redaction` today. I verified each by running it rather than reading it:

**Connection-URI userinfo (P1)** — #207 adds a DSN pattern. Run against its `scrubSecrets`:

```
mongodb://user:hunter2@host/db          -> mongodb://user:[REDACTED]@host/db
postgres://admin:hunter2@db:5432/app    -> postgres://admin:[REDACTED]@db:5432/app
mysql://root:p%40ss@10.0.0.1:3306/x     -> mysql://root:[REDACTED]@10.0.0.1:3306/x
```

**Bare CLI arguments (P2)** — #207 fixes it at the producer, which is the stronger form of what you suggest: `agent_cli.ts` no longer builds the token into the message at all (`unrecognized argument (expected <flag list>)`), so there is nothing for a scrub to miss.

**Why I am pointing rather than duplicating.** #207 rewrites the same regions of `connector.ts` with a deeper redaction pass. Merged after it, this branch produced 9 conflict hunks, one spanning ~570 lines. I removed the Prettier reformat in `1b8fe60` to bring that to 4 small hunks; re-implementing URI redaction and the CLI diagnostic here would put it straight back, for behavior that lands either way.

### Fix criterion for the rest of this cycle

I will fix a finding here when it is (a) this branch's own regression, and (b) testable with one right answer — the item above was both. A finding whose remedy is already implemented on #207 gets verified and pointed at, not re-implemented, because duplicating it costs the maintainer merge conflicts and buys nothing.

### Merge note

`#207` then `#227` is the order I verified. The 4 remaining `connector.ts` hunks all have #207's side as a strict superset — **take HEAD at every one**. In `connector.test.ts`, take HEAD *and* keep this branch's `scrubbing the override is idempotent` case, which #207 does not have. Resolved that way and run: **2988 passing, tsc clean**.

rfv-code:
## Round 3 — both fixed in `f8b8de6`. One of them was a sibling I should have swept for.

Full suite **2810 passing**, tsc clean. Threads resolved.

### 1. Truncation defeated the scrub (P1)

Confirmed, and measured rather than reasoned about. `scrubSecrets` matches a quoted value by finding its **closing** quote, and the 200-character cut lands inside a long one. A JWT in a 4xx body:

```
scrub -> truncate : {"error":"bad","token":"[REDACTED]"}
truncate -> scrub : {"error":"bad","token":"eyJhbGciOiJIUzI1NiJ9.AAAAAAAA…
```

The second is what shipped, on every HTTP connector. Scrubbing first is also strictly cheaper, since the value is a short marker by the time truncation runs.

### 2. Aborts classified from the scrubbed message (P2) — my miss

Confirmed. This is the **sibling** of the `defaultErrorMap` fix I pushed in `014ae04` one commit earlier, and I did not sweep for it. Writing the general rule is not the same as applying it at every call site, and that is the honest description of what happened here.

I have now swept both classes rather than fixing only the reported instance:

- **A decision derived from a scrubbed string** — exactly two sites in `src/`: `connector.ts:240` (fixed in `014ae04`) and `http_client.ts:314` (fixed here). No third.
- **Truncation of a string that a scrub must later match** — one site, fixed here. The only other length cap in `src/` is `gcp/lib/gcp_client.ts:594`, which truncates a single line it does not hand onward to a shape-based scrub.

Note the consequence differs from item 1: this one is a wrong error code, not a leak. A flattened `Authorization: … aborted` scrubs to `Authorization: [REDACTED]`, the abort marker goes with it, and the request is reported `NETWORK_ERROR` instead of the documented `TIMEOUT` — which, as you point out, the connector tests rely on.

### Verification

3 tests, each verified to fail against its own mutation, including one pinning that a non-abort failure is *still* scrubbed on the way out — so the classification fix cannot be mistaken for weakening the redaction.

### Where this cycle stands

The criterion I posted last round still holds and I have paid it twice now. Worth naming the trend though: of the five findings across rounds 2 and 3, two were consequences of my own patches from the round before. That is the shape where a review loop stops being about the PR. I will keep fixing findings that are this branch's own regressions and testable with one right answer, and I will keep pointing at #207 for anything already implemented there rather than duplicating it into a merge conflict.

The merge note from round 2 is unchanged: `#207` then `#227`, take HEAD at every `connector.ts` hunk, and keep this branch's idempotence case in `connector.test.ts`.

rfv-code:
## Round 4 — fixed in `6eca62f`, plus its sibling. Thread resolved.

Full suite **2812 passing**, tsc clean, CI green.

### Confirmed, and the measurement is worth having

`JSON.parse` quotes the offending source back, but not always in full. Measured on the Node this repo runs:

| source length | parser message contains |
|---|---|
| ≤ 18 chars | the input **in full** |
| > 18 chars | a **10-character prefix** plus `...` |

So `--params hunter2` wrote `hunter2` to the envelope on stdout *and* to stderr. Both shapes are leaks — plenty of real credentials are under 18 characters, and a 10-character prefix of a longer one is still a partial disclosure.

### The sibling, which the report did not cover

`agent_cli.ts` quoted a rejected positional or flag name straight back (`unrecognized argument: ${a}`) — the same leak by a different route. You had flagged the positional case separately in round 2 and I pointed at #207 rather than fixing it; measuring the class this time, both sites are the same defect and both are now fixed. Both name the accepted flags instead of the rejected token.

### A correction to what I said last round

I claimed in the commit message that this "does not add merge surface." **That was wrong** — I checked afterwards. Touching `agent_cli.ts` adds two conflicting files against #207:

| file | hunks before | after |
|---|---|---|
| `connector.ts` | 5 | 5 |
| `connector.test.ts` | 1 | 2 |
| `agent_cli.ts` | 0 | **2** |
| `agent_cli.test.ts` | 0 | **1** |

The new `agent_cli.ts` hunks are **comment-only** — both branches implement the identical rule and differ in how the comment is worded — so they cost nothing to resolve. I judged the live leak worth that, since #227 may merge first.

### Updated merge rule (verified, not guessed)

`#207` first, then `#227`:

- `connector.ts`, `agent_cli.ts`, `.jules/sentinel.md` — **take HEAD** at every hunk.
- `agent_cli.test.ts` — take HEAD, then add #227's `throws on a positional argument without echoing it`. HEAD pins the exact message; only #227's asserts the token is absent.
- `connector.test.ts` — take HEAD, then add #227's `scrubbing the override is idempotent` and `classifies from the raw message`. HEAD's `a malformed --params never echoes the input, whatever its shape` is **broader** than #227's equivalent (three shapes including a DSN), so #227's `--params` test is dropped rather than kept.

Resolved exactly that way and run: **2996 passing, tsc clean, zero unmerged paths.**

Note the classification test holds by construction on #207 — its `defaultErrorMap` already reads the raw message — so it is a regression pin there rather than a fix. Worth keeping for that reason.

rfv-code:
## Round 5 — confirmed, and it is already implemented on #207. Not duplicating it here.

Reproduced exactly as described. A credential in an enum field, on **this** branch:

```
state_reason: Invalid enum value. Expected 'completed' | 'not_planned', received 'ghp_AAAABBBB'
```

The token is in the envelope. `scrubSecrets` cannot see it — there is no key beside it, and shape is all that function has.

On `fix/connector-credential-redaction`, same connector, same input:

```
state_reason: Invalid enum value. Expected 'completed' | 'not_planned', received '[REDACTED]'
```

That is your remedy, already built: #207 collects the caller's own values as redaction candidates and decides per issue whether the path is sensitive, rather than scrubbing Zod's rendered string. It also keeps the useful half — the accepted enum values survive, which a message-level drop would have cost.

### Why I am not implementing it here

The remedy is not a small change. On #207 it is the candidate collector, `isSensitiveFieldPath`, `redactEchoedInput`, and a comparison budget — a large body of code that has taken most of that PR's 28 rounds to get right, and it lives in the same regions of `connector.ts` this branch already conflicts with. Duplicating it would recreate exactly the merge surface I spent two commits reducing, for behavior that lands either way.

This is the criterion I posted in round 2 and it has now cut both ways: I fixed the `--params` echo here last round because #227 could merge first and the fix converged with #207's rather than diverging; this one cannot converge, so it gets verified and pointed at instead.

**If #227 merges before #207, this leak ships.** It is a P2 by your rating and I agree with that — the value has to be supplied by the caller to an enum field. The merge order I have verified and posted twice is `#207` then `#227`, which closes it.

Merge rule unchanged from round 4.

rfv-code (commented):
## Verdict: approve — merge **after #207**, or drop the two files it duplicates

Reviewed at tip `6eca62f`. I re-derived the three premises this PR rests on by running `scrubSecrets` rather than reading the regexes, and all three hold. The scope that is genuinely this PR's own — `http_client.ts`, `cli/usage-report.ts`, `_optional.ts` — is correct and covered by tests. The `connector.ts` and `agent_cli.ts` hunks are a subset of #207 and are the only thing standing between these two PRs and a clean merge.

---

### 🟢 Good — the premises check out under execution

**1. Classify from the raw, return the scrubbed** (`connector.ts:239-241`, `http_client.ts:326-328`)

This is the load-bearing change and the reasoning in the comments is right. Measured:

```
in : Authorization: Bearer abc.def.ghi 401 Unauthorized
out: Authorization: Bearer [REDACTED]
```

`SENSITIVE_AUTH_LINE_RE` runs to end-of-line, so scrubbing first takes the `401` and the `unauthor` with it. `defaultErrorMap` would then fall through to `CONNECTION_ERROR`, which is in `RETRIABLE_CODES` (`connector.ts:237`) — an auth failure silently retried as a network blip. Same for the abort marker in `http_client.ts`: `"...Bearer abc.def.ghi request aborted"` scrubs to `"Authorization: Bearer [REDACTED]"`, `/abort/i` stops matching, and a documented `TIMEOUT` is reported as `NETWORK_ERROR`. Both restructures are necessary, not stylistic.

**2. Scrub before truncate** (`http_client.ts:283-289`)

Correct, and the stated mechanism is the real one. `scrubSecrets` locates a quoted value by finding its *closing* quote, so a 200-char cut landing mid-JWT leaves an unterminated value that no pattern matches, and the prefix survives into the envelope. Scrubbing first collapses the value to a short marker:

```
in : {"token":"eyJhbGciOiJIUzI1NiJ9.payload.sig"}
out: {"token":"[REDACTED]"}
```

**3. The `--params` position-digits rule** (`connector.ts:599-614`)

The premise is not hypothetical — I confirmed the parser message passes through untouched:

```
in : Unexpected token g, "ghp_AAAABBBBCCCC" is not valid JSON
out: Unexpected token g, "ghp_AAAABBBBCCCC" is not valid JSON     <- unchanged
```

A bare token has no `key = value` shape, so `scrubSecrets` cannot see it. Copying only the position digits is the right general form, and dropping the detail when Node omits the position is the right fallback.

**4. Two claims made in comments, both verified**

- *`scrubSecrets` is idempotent* (`connector.ts:755`) — held on all 9 inputs I probed, including the AUTH-line and quoted-value branches.
- *scrubbing the `mapError` override covers all nine mappers* (`connector.ts:749-756`) — `cfg.mapError` has exactly one consumer, `mapAndBuildError` at `connector.ts:746`. Nothing else calls it.

**5. No import cycle.** `_optional.ts:1` and `http_client.ts:1` now import from `audit/writer.js`, which imports only `node:crypto`, `node:fs`, `node:path` and a type from `./events.js`. Safe.

---

### 🟠 Major — `connector.ts` and `agent_cli.ts` duplicate #207

Not a correctness defect; a merge-order problem that will cost someone an afternoon if it is discovered at merge time.

`agent_cli.ts` is the clearer case: **both PRs add a character-identical `flagList` helper and the same two message strings.** Only the comment wording differs. This is one derivation done twice.

`connector.ts` is a strict subset of #207 on every shared hunk:

| hunk | this PR | #207 |
|---|---|---|
| `policyLoadError` | `scrubSecrets(...)` | same, plus a DO-NOT-REMOVE pin |
| `classify() threw` | `scrubSecrets(...)` | same, then `redactSensitiveEchoes(...)` against the caller's params |
| `extendDecision() threw` | `scrubSecrets(...)` | same, plus the sibling-hook note |
| `--params` JSON | position digits only | same rule, same reasoning |
| `mapError` override | `scrubSecrets(override.message)` | scrubs at the single sink, covering **both** branches |

On the one hunk where the two differ in kind — `defaultErrorMap` — #207 is also correct: it never scrubs inside the mapper, so `lower` is already the raw string there, and the single `scrubSecrets` in `mapAndBuildError` covers the default and override branches together. That is the better shape of the same fix.

**Measured merge result.** Merging #207 then this PR onto `main` produces 5 conflicts:

```
.jules/sentinel.md
src/toolkit/agent_cli.ts
src/toolkit/connector.ts
tests/toolkit/agent_cli.test.ts
tests/toolkit/connector.test.ts
```

**Verified resolution — take #207's side on all five** (keep both notes in `sentinel.md`). With that resolution: `tsc --noEmit` exit 0, `vitest run` **2999 passed / 0 failed**. Nothing from this PR's exclusive scope is lost — `http_client.ts`, `cli/usage-report.ts` and `_optional.ts` do not conflict at all.

I also merged the full open queue (#207, #227, #210, #94, #111, #112, #106) onto `main` in one tree using that rule: typecheck clean, **3035 passed / 0 failed**. The whole queue is compatible.

Recommendation, either is fine:

- **Merge #207 first**, then this PR with the resolution above; or
- **Drop `src/toolkit/connector.ts` and `src/toolkit/agent_cli.ts` from this PR** so it carries only its exclusive scope and merges in any order.

---

### 🟡 Minor

1. **`_optional.ts:1`, `http_client.ts:1` — import above the module docblock.** The new `import` was inserted before the file's own header comment, so the docblock now documents the import rather than the module, and JSDoc tooling will attach it there. Move the import below the closing `*/`.

2. **Unrelated reformat hunks.** `http_client.ts` reformats `_connectTimeoutMs` (155), `filename:` (309), `_buildUrl` (374) and `headerHasCT` (406); `usage-report.ts` restyles the whole `parseArgs` if-chain. None relate to scrubbing. Harmless, but they are most of the deletion counts and make the security diff harder to read than it is.

Both are optional polish. I did not push a fix for either, deliberately: this PR is green and quiet after five review rounds, and a cosmetic push would restart the cycle for no behavioral gain.

---

### 🟡 Residue — named, not fixed

1. **`fetch_attachment.ts:155,180`** still place a raw `e.message` into the `warning` field of a *success* envelope. These come from `extractBinary` operating on a temp file under `os.tmpdir()` with a `randomUUID` name, so what escapes is a parse error and a random path — not a credential. Out of this PR's class; flagging so it is a decision rather than an oversight.

2. **A connection URL with inline userinfo is not redacted on this branch:**

```
in : postgres://user:pw@host/db
out: postgres://user:pw@host/db     <- unchanged
```

That matters because the DB drivers relay vendor SDK messages verbatim and `src/connectors/db/connector.ts:175` returns `err.message` through `mapError`. **This is not a gap to fix here** — #207 adds the userinfo rule to `writer.ts` (along with PEM blocks, camelCase keys and Digest params), so it closes when #207 lands. Recording it because it is the one credential shape this PR's own title would lead you to expect it covers.

---

### Punch list

| # | Sev | Item | File:line | Action |
|---|-----|------|-----------|--------|
| 1 | 🟠 | `connector.ts` + `agent_cli.ts` duplicate #207; 5 merge conflicts | `connector.ts`, `agent_cli.ts` | Merge #207 first (take its side), or drop both files here |
| 2 | 🟡 | Import placed above the module docblock | `_optional.ts:1`, `http_client.ts:1` | Move below the docblock — optional |
| 3 | 🟡 | Unrelated Prettier hunks inflate the diff | `http_client.ts:155,309,374,406` | Optional |
| 4 | 🟡 | Raw `e.message` in a success envelope's `warning` | `fetch_attachment.ts:155,180` | Out of class — decide separately |
| 5 | 🟢 | URL-userinfo credentials unredacted | `audit/writer.ts` | Closes with #207 — no action here |

**CI:** `test (22)` and `test (24)` both green. **Threads:** 9, all resolved. No open findings against this PR's own scope.

rfv-code:
Closing as a duplicate of #207, with this PR's exclusive scope landed there first (01fb31d).

**Why #207 is canonical.** Both PRs redact credentials on connector error paths, and they conflict in `src/toolkit/connector.ts`, `src/toolkit/agent_cli.ts` and their tests. #207 is the broader and older line of work: 64 commits, 112 resolved review threads, and it supersedes #154. Where this PR adds `scrubSecrets` at seven separate sites in `connector.ts`, #207 scrubs once at a choke point in `mapAndBuildError` that covers the `cfg.mapError` override branch and the `defaultErrorMap` branch together, then applies `redactSensitiveEchoes` — input-aware redaction that this PR does not have. The `flagList` change in `agent_cli.ts` is present, near-identical, in both.

**What was ported (now on #207 as 01fb31d).** `src/toolkit/cli/usage-report.ts`. It is a standalone CLI whose `parseArgs` echoed the caller's rejected token to stderr, outside the connector envelope, so the choke-point scrub never saw it. Its two test cases came across as well, with the rejected-argument assertion tightened: it now pins the message text rather than a flag name, because HELP is written to the same stream and already names every flag, so the original assertion passed with the fix removed. Both cases fail against a mutated build. Full suite: 2996 passed, 0 failed.

**What was deliberately not ported, and why.**

1. `src/toolkit/_optional.ts` — retracting this one. The change scrubs `msg`, but `msg` is classification-only: it appears solely in the four `msg.includes(...)` tests that decide `notFound`, and it never reaches output. The thrown message is built from `pkgName`, derived from `modName`/`installName`, and all five call sites pass string literals. So it redacts an input to a predicate, which is the ordering error this PR's own `connector.ts` comment warns against, and it fixes no leak. The unhandled path in that function is `throw e` on the non-not-found branch, which this change does not touch.

2. `http_client.ts` `mapHttpError` — redundant on #207. All six `mapHttpError(CODE_MAP)` consumers reach output through `mapAndBuildError`, which scrubs at `connector.ts` after the override branch.

3. `http_client.ts` scrub-before-truncate — the stated mechanism does not reproduce. The rationale is that a 200-character cut lands inside a quoted value, the closing quote is lost, and the prefix survives. I ran both orderings over eight payload shapes (JSON `access_token`, JSON `password`, flattened `Authorization` header, `api_key=`, nested JSON, query string, YAML-like) with a 316-character secret. `scrubSecrets` matched the unterminated value in every one: zero leaks under truncate-first. Scrub-first is still the better order for message quality — the redaction marker is short, so less diagnostic text is cut — but that is a readability argument, not the security one given.

4. `http_client.ts` raw-vs-scrubbed abort classification — **this one is a real finding and is not yet fixed anywhere.** It reproduces: `scrubSecrets('Authorization: Bearer sk-abc123 request aborted')` returns `'Authorization: Bearer [REDACTED]'`, so `/abort/i` matches the raw string and not the scrubbed one. Any fix that scrubs before classifying turns a TIMEOUT into a NETWORK_ERROR. Separately, `http_client.ts:318` returns the raw `err.message` on the non-aborted branch with no scrub at all. I did not port it here because it belongs to `src/toolkit/http_client.ts`, a file #207 does not touch, and folding a new file into a 64-commit PR that is otherwise ready is the wrong trade. It should be its own PR: scrub `err.message` and the `bodyText` in the HTTP-status message, keeping `/abort/i` on the raw string.

PR #228, opened 2026-09-08
⚡ Bolt: Remove redundant sorting of statement boundaries

rfv-code:
💡 **What:** 
Replaced `Set<number>` with `number[]` in `_boundarySemicolons` and removed the `Array.from(...).sort(...)` call in `_splitStatements`.

🎯 **Why:** 
The SQL string is parsed sequentially, so boundary indices are discovered in strictly increasing order. Dumping them into a `Set` and sorting them explicitly adds unnecessary O(N log N) overhead, which can be noticeably slower on extremely large scripts. This leverages the natural sequence of the `push()` operations.

📊 **Impact:** 
Changes statement boundary verification time complexity from O(N log N) to O(N).

🔬 **Measurement:** 
Run `pnpm test tests/connectors/db/policy.test.ts`. All 59 tests should pass successfully.

---
*PR created automatically by Jules for task [2314840377359573736](https://jules.google.com/task/2314840377359573736) started by @rfv-code*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -10,3 +10,6 @@
 ## 2024-06-25 - Avoid array sort for Top-K in hot paths
 **Learning:** V8 engine sorting (`[...arr].sort()`) is surprisingly slow and blocks the event loop for thousands of records when fetching Top-K elements (e.g. usage statistics). It scales at O(N log N) when O(N) is sufficient for a fixed K.
 **Action:** When gathering top elements from a large data array in this codebase, manually track the top items in a single O(N) pass rather than sorting a full copy.
+## 2025-03-08 - O(N log N) Set to Array sorting is unnecessary when iteration naturally provides order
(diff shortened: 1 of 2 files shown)


rfv-code:
Closing as duplicate of #111.

Both PRs make the identical change in `src/connectors/db/lib/policy.ts`: `_boundarySemicolons` returns `number[]` instead of `Set<number>`, uses `push` instead of `add`, and `_splitStatements` drops the `Array.from(...).sort((a, b) => a - b)` round-trip and compares lengths instead of `Set.size`.

#111 is canonical because it ships tests for the behavior — `tests/connectors/db/policy_security.test.ts` gains a case pinning the parity check at the *position* level (a boundary disagreement where both escape modes find the same count but at different offsets, which a length-only comparison would let through) and a case pinning ordered multi-statement splitting. This PR has no tests.

The only content unique to this PR is a whole-file Prettier reformat, which is unrelated churn.

If you believe some case here is not covered by #111, please cherry-pick it onto that branch.

PR #229, opened 2026-09-09
⚡ Bolt: Optimize SQL statement boundary parsing

rfv-code:
💡 What: Replaced `Set<number>` with `number[]` and removed the redundant sorting in SQL boundary parsing (`src/connectors/db/lib/policy.ts`).
🎯 Why: Because the string is parsed sequentially, the indices of semicolons are discovered in strictly increasing order. Adding them to a `Set` and then sorting the resulting array introduces unnecessary O(N log N) overhead and memory allocations. A simple array with `.push()` does the job in O(N).
📊 Impact: Eliminates an unnecessary `Set` allocation and array sort operation per statement block being evaluated by the policy engine.
🔬 Measurement: Verified with `pnpm test` (all 2800+ tests pass) to ensure accuracy in statement boundaries is preserved while parsing.

---
*PR created automatically by Jules for task [364123110867909998](https://jules.google.com/task/364123110867909998) started by @rfv-code*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -10,3 +10,6 @@
 ## 2024-06-25 - Avoid array sort for Top-K in hot paths
 **Learning:** V8 engine sorting (`[...arr].sort()`) is surprisingly slow and blocks the event loop for thousands of records when fetching Top-K elements (e.g. usage statistics). It scales at O(N log N) when O(N) is sufficient for a fixed K.
 **Action:** When gathering top elements from a large data array in this codebase, manually track the top items in a single O(N) pass rather than sorting a full copy.
+## 2024-05-26 - SQL Statement Boundary Parsing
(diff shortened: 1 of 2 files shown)


rfv-code:
Closing as duplicate of #111.

The only behavioral change here is the one #111 already makes: `_boundarySemicolons` returns `number[]` instead of `Set<number>`, and `_splitStatements` compares the two arrays positionally instead of `Array.from(...).sort()`. Everything else in this diff (+108/-47 vs #111 at +9/-9) is a Prettier reformat of lines neither PR is changing — the keyword `Set` literals, `classifySqlKeywords` if-bodies, the `_splitStatements` signature.

#111 also ships `tests/connectors/db/policy_security.test.ts` (+23) covering the ordering invariant the optimization depends on. This PR ships no test.

Diffed against the merge base: zero residual behavior here over #111.

PR #230, opened 2026-09-09
🛡️ Sentinel: [CRITICAL] Fix Slowloris DoS vulnerability in fetch timeouts

rfv-code:
🚨 **Severity:** CRITICAL
💡 **Vulnerability:** When implementing timeouts for streaming fetch requests, `clearTimeout()` was called immediately after the initial fetch (header) resolved, before the response body stream was read. This allowed a malicious server to drip the response body indefinitely, hanging the process (a Slowloris-style DoS attack).
🎯 **Impact:** An attacker could craft a slow HTTP response that hangs the connector indefinitely, consuming resources and eventually causing a denial of service (DoS) across the agent infrastructure.
🔧 **Fix:** Expanded the `try...finally` block in `src/toolkit/fetch_helper.ts` and `src/toolkit/fetch_attachment.ts` to ensure `clearTimeout()` only executes *after* the entire body stream is read or buffered, correctly interrupting slow streams.
✅ **Verification:** Verified by running the relevant test suite for the HTTP toolkit (`pnpm test tests/toolkit/fetch_helper.test.ts tests/toolkit/fetch_attachment.test.ts tests/toolkit/http_client.test.ts`). Added an entry to `.jules/sentinel.md` noting the vulnerability and the prevention pattern.

---
*PR created automatically by Jules for task [6306984650818859148](https://jules.google.com/task/6306984650818859148) started by @rfv-code*


rfv-code:
Closing as duplicate of #112.

Both move the body read inside the `try`/`finally` in `fetchWithCaps` so the timeout timer stays armed through the stream — that is the whole of this PR. #112 has that same restructure plus scope this PR lacks:

- `fetch_attachment.ts` delegates to `fetchWithCaps` instead of calling `doFetch` directly, so it gets the **streaming** byte cap, not just the `content-length` header check. This PR leaves `fetchAttachment` on its own raw fetch, so a chunked response with no `content-length` still buffers unbounded there.
- The timer aborts with `new DOMException("fetch_helper timeout", "AbortError")` rather than a plain `Error`, so the conventional `err.name === "AbortError"` check that both docblocks advertise actually works.
- `fetchImpl` is threaded through as a parameter (needed to test the above).
- `tests/toolkit/fetch_helper.test.ts` (+85 lines). This PR ships no tests.

Diffed against the merge base per file: zero residual scope here over #112. If some edge case is missed, cherry-pick it onto `sentinel/fix-dos-fetch-attachment-16824957574873726734`.

PR #231, opened 2026-09-10
⚡ Bolt: Remove redundant sort and Set conversion in SQL boundary parsing

rfv-code:
💡 What: Refactored `_boundarySemicolons` in `src/connectors/db/lib/policy.ts` to accumulate indices in a standard array (`number[]`) instead of a `Set<number>`, avoiding the subsequent `Array.from()` and `.sort()` operations in `_splitStatements`.
🎯 Why: Because the string is iterated sequentially from left to right, the collected indices are naturally strictly increasing. Converting to a Set and then sorting them introduced unnecessary O(N log N) overhead and memory allocations for large compound SQL strings.
📊 Impact: Eliminates O(N log N) sorting overhead and Set object allocations on every database query parsed by the policy engine, making the hot-path `_splitStatements` function strictly O(N).
🔬 Measurement: Verify tests run successfully using `pnpm test tests/connectors/db/policy.test.ts`. Inspect the policy module to ensure the arrays remain correctly monotonically increasing.

---
*PR created automatically by Jules for task [2623604652305352687](https://jules.google.com/task/2623604652305352687) started by @rfv-code*

### DIFF
--- a/.jules/bolt.md
+++ b/.jules/bolt.md
@@ -10,3 +10,6 @@
 ## 2024-06-25 - Avoid array sort for Top-K in hot paths
 **Learning:** V8 engine sorting (`[...arr].sort()`) is surprisingly slow and blocks the event loop for thousands of records when fetching Top-K elements (e.g. usage statistics). It scales at O(N log N) when O(N) is sufficient for a fixed K.
 **Action:** When gathering top elements from a large data array in this codebase, manually track the top items in a single O(N) pass rather than sorting a full copy.
+## 2024-06-25 - SQL Boundary Parsing Redundant Sort Removed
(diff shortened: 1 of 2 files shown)


rfv-code:
Closing as a duplicate of #111.

Same change: `_boundarySemicolons` returns `number[]` instead of `Set<number>`, `.add` becomes `.push`, `.size` becomes `.length`, and `_splitStatements` drops the `Array.from(...).sort(...)` pair.

I merged both branches onto current main and diffed the resulting `src/connectors/db/lib/policy.ts` with `diff -w -B`. Every remaining difference is Prettier line-wrapping or the wording of the `⚡ Bolt` comment — there is no functional residue in this PR at all. What #111 has in addition is 23 lines of coverage in `tests/connectors/db/policy_security.test.ts`; this branch has none, and it also carries a whole-file reformat that would collide with #94, which is queued against the same file.

Nothing to cherry-pick onto #111.

PR #233, opened 2026-09-11
🛡️ Sentinel: [HIGH] Fix Slowloris DoS vulnerability in fetchWithCaps

rfv-code:
**🚨 Severity:** HIGH
**💡 Vulnerability:** A malicious server could intentionally stall an HTTP response body (trickle feeding) after sending headers. Because `clearTimeout(timer)` was called immediately after the initial fetch resolved, the timeout was bypassed, leading to potential exhaustion of application resources while waiting on a stalled stream indefinitely (Slowloris-style DoS).
**🎯 Impact:** A targeted denial of service where limited client network connections or memory could be indefinitely locked up by a malicious external server.
**🔧 Fix:** Expanded the scope of the `try...finally` block that clears the timer so it covers the entire request lifecycle, including the `reader.read()` loop that consumes the response body.
**✅ Verification:** Added a unit test in `fetch_helper.test.ts` that mocks a `ReadableStream` which stalls indefinitely after the first chunk, asserting that the timeout correctly fires and aborts the request.

---
*PR created automatically by Jules for task [683931236545960516](https://jules.google.com/task/683931236545960516) started by @rfv-code*


rfv-code:
Closing as duplicate of #112. #112 makes the identical Slowloris fix (moves the body-read loop inside the `try` so `clearTimeout` only runs in the `finally` after the body is drained) and additionally: aborts with a `DOMException` named `AbortError` so callers' conventional `err.name === "AbortError"` check classifies a body-phase timeout correctly, injects `fetchImpl` for testability, and applies the same caps to `fetch_attachment.ts`. Its test file covers this PR's trickle-feed case ("keeps the timeout armed during the body stream") plus three more. Nothing here is missing from #112.

PR #240, opened 2026-09-15
⚡ Bolt: Optimize SQL statement boundary tracking (O(N log N) -> O(N))

rfv-code:
💡 What: Replaced a `Set` with a standard `number[]` array to track sequentially discovered statement boundaries in `src/connectors/db/lib/policy.ts`, explicitly avoiding a redundant `Array.from(...).sort()` call.
🎯 Why: Because the string parsing loop increments indices sequentially, the identified boundaries are naturally perfectly sorted and have no duplicates. The O(N log N) sorting added overhead without any benefit. 
📊 Impact: Completely eliminates O(N log N) overhead and object allocation overhead (via Set) for parsing SQL boundaries. The process is now linear O(N) allocation and lookup. 
🔬 Measurement: Run `pnpm test tests/connectors/db/policy.test.ts` to verify boundaries logic continues to accurately break up statements without regressions.

---
*PR created automatically by Jules for task [2466130615596831747](https://jules.google.com/task/2466130615596831747) started by @rfv-code*


rfv-code:
Closing as duplicate of #111.

#111 makes the identical change to the identical function and adds the regression tests this PR lacks. I compared the two tips token-by-token (comments and whitespace stripped) on `src/connectors/db/lib/policy.ts`: **every** difference is Prettier formatting — added trailing commas, re-wrapped `if` conditions, reflowed `Set([...])` literals. There is no semantic residual.

Both PRs make exactly this change:
- `_boundarySemicolons` returns `number[]` instead of `Set<number>` (`.add` -> `.push`)
- `_splitStatements` compares `.length` instead of `.size` and drops both `Array.from(...).sort((a,b)=>a-b)` calls

What #111 adds that this PR does not:
- `rejects a boundary disagreement even when both modes find the same count` — pins the parity check at the *position* level. Dropping the sort is only safe because the scan is a single forward pass; a count-only comparison would let `SELECT 'a\\';x'; SELECT 2` split at the attacker-chosen offset.
- `splits a multi-statement script at every boundary in order` — pins the ascending, duplicate-free ordering the removed `Set` used to guarantee.

Since this is a security-relevant parser, the optimization should not land without those two tests. The whole-file reformat here would also collide with #94, which is in flight on the same file.

If you want the formatting normalized, that belongs in its own PR.

PR #242, opened 2026-09-16
🛡️ Sentinel: [CRITICAL] Fix Slowloris DoS vulnerability in fetchWithCaps

rfv-code:
🚨 Severity: CRITICAL
💡 Vulnerability: The `fetchWithCaps` function previously called `clearTimeout(timer)` immediately after the initial HTTP `fetch` resolved. Because `fetch` resolves as soon as the HTTP headers are received, the timeout did not apply to the subsequent streaming of the response body.
🎯 Impact: A malicious server could exploit this by returning headers immediately and then trickling the response body extremely slowly (a Slowloris attack). This would bypass the fetch timeout, causing the `fetchWithCaps` function to hang indefinitely and exhaust system resources.
🔧 Fix: Moved the `clearTimeout(timer)` call out of the initial `fetch` block and placed it in a `finally` block that wraps the entire body streaming and parsing loop. This ensures the configured timeout applies to the full request lifecycle.
✅ Verification: Added a new test `aborts when the response body trickles too slowly (Slowloris)` to `tests/toolkit/fetch_helper.test.ts` that mocks a slow `reader.read()` stream. Verified that `fetchWithCaps` correctly throws a timeout error and that the entire suite passes.

---
*PR created automatically by Jules for task [86754907231530780](https://jules.google.com/task/86754907231530780) started by @rfv-code*


rfv-code:
Closing as duplicate of #112, which makes the same `try`/`finally` move in `fetchWithCaps` (timer stays armed through the body read), also routes `fetchAttachment` through it, and already has the equivalent Slowloris regression test (`keeps the timeout armed during the body stream`) plus `AbortError` tests in `tests/toolkit/fetch_helper.test.ts`. This PR differs from main + #112 in one way, and it is a regression: the timeout aborts with a plain `Error`, not the `DOMException`/`AbortError` that `fetchWithCaps`'s docblock promises (main `src/toolkit/fetch_helper.ts:74`). Measured with a trickling-body fetch stub, timeout 100 ms:

| implementation | result |
|---|---|
| main | still reading after 1500 ms (no body timeout) |
| this PR | rejects in ~115 ms, `Error name=Error` |
| #112 | rejects in ~116 ms, `DOMException name=AbortError` |

PR #246, opened 2026-09-18
🛡️ Sentinel: [HIGH] Fix Slowloris DoS vulnerability in fetchWithCaps

rfv-code:
🚨 Severity: HIGH
💡 Vulnerability: A Slowloris-style denial of service vulnerability in `fetchWithCaps` where the stream reading loop was unbounded by timeout because `clearTimeout(timer)` was called immediately after the initial fetch resolved.
🎯 Impact: An attacker or malicious server could trickle response bodies indefinitely, tying up application resources, memory, and keeping open connections alive far beyond the intended `timeoutMs` threshold.
🔧 Fix: Wrapped the initial HTTP headers fetch and the entire `response.body?.getReader().read()` loop inside a single `try/finally` block. `clearTimeout(timer)` is now correctly invoked *after* the entire stream is fully processed and bounded.
✅ Verification: Ran `pnpm test tests/toolkit/fetch_helper.test.ts` to ensure standard caps, bounds, and normal fetching still behave correctly.

---
*PR created automatically by Jules for task [5873325453231849990](https://jules.google.com/task/5873325453231849990) started by @rfv-code*

PR #247, opened 2026-09-19
🛡️ Sentinel: [HIGH] Fix Slowloris DoS vulnerability in fetch helpers

rfv-code:
🚨 **Severity:** HIGH
💡 **Vulnerability:** `fetchWithCaps` and `fetchAttachment` were clearing their abort timeout timers immediately after headers arrived (i.e. right after `fetch` resolved), leaving the subsequent stream processing operations unprotected against Slowloris-style Denial of Service attacks.
🎯 **Impact:** A malicious or degraded server could send headers to satisfy the initial `fetch` timeout and then indefinitely trickle response body chunks at a tiny rate, holding connections open and eventually exhausting application threads/memory (DoS).
🔧 **Fix:** Expanded the scope of the `try...finally { clearTimeout(timer) }` blocks in both `fetch_helper.ts` and `fetch_attachment.ts` to encapsulate the entire body reading process (`reader.read()` and `response.arrayBuffer()`), ensuring the network timeout protects the full request lifecycle.
✅ **Verification:** Verified by implementing a local HTTP server that responds with valid headers but never terminates the chunked stream, confirming the abort signal now properly terminates the pending body streams. Tests for `fetch_helper` and `fetch_attachment` continue to pass.

---
*PR created automatically by Jules for task [5121224217734851782](https://jules.google.com/task/5121224217734851782) started by @rfv-code*


rfv-code:
Closing as duplicate of #112.

#112 is the canonical Slowloris fix: 7 commits, 85 lines of new tests in `tests/toolkit/fetch_helper.test.ts`. This PR is one commit with no tests, and its change is a strict subset of #112's on both files.

`src/toolkit/fetch_helper.ts` — this PR's version is **byte-identical to #246's**, and both are #112's `f6718d3` (move the body-stream loop inside the `try` so `clearTimeout` only fires in the `finally`) *without* #112's follow-up `8860521`. That follow-up matters: #112 aborts with `new DOMException("fetch_helper timeout", "AbortError")`, this PR keeps `new Error("fetch_helper timeout")`. The reason propagates verbatim to whatever `fetch`/`reader.read()` rejects with, so the conventional `err.name === "AbortError"` check misclassifies a timeout as a generic network failure — on two paths now that the timer stays armed through the body. #112 also adds the `fetchImpl` parameter, which is what makes the timeout testable at all.

`src/toolkit/fetch_attachment.ts` — the two PRs diverge here, and #112's approach is the safer one. This PR keeps `fetchAttachment`'s private copy of the fetch/timeout/cap logic and moves `arrayBuffer()` inside the `try`. That closes the header-then-drip window, but the body is still fully buffered before `rawBytes.byteLength > maxBytes` is checked, so a response that under-reports or omits `content-length` still allocates the whole payload. #112 deletes that private copy and delegates to `fetchWithCaps(url, init, { maxBytes, timeoutMs }, opts.fetchImpl)`, which enforces the cap **incrementally during the stream** and cancels the reader as soon as `total > maxBytes`. One implementation, one place to fix.

**0 declarations exclusive to this PR.** If you think a case here is missing from #112, please cherry-pick it onto that branch.

PR #248, opened 2026-09-19
⚡ Bolt: Replace O(N log N) sorting for statement boundary splitting with O(N) array

rfv-code:
💡 What: Replaced the use of `Set<number>` for collecting statement boundary indices with a standard `number[]` array in `_boundarySemicolons`, completely removing the need to convert and sort it with `Array.from(set).sort()` inside `_splitStatements`.

🎯 Why: In `_boundarySemicolons`, indices are naturally encountered and pushed in monotonically increasing order since the loop iterates forward (`i++`) through the SQL string. Constructing a `Set` only to cast it back to an Array and run V8's sorting algorithm introduced unnecessary $O(N \log N)$ overhead that scales poorly with very large or complex SQL statements.

📊 Impact: Reduces the time complexity of compiling statement boundary metrics from $O(N \log N)$ down to a flat $O(N)$, saving CPU cycles and minimizing main-thread blocking during intense parsing of vast SQL strings.

🔬 Measurement: Verified functionality and syntax preservation using `pnpm test tests/connectors/db/policy.test.ts`. There are zero behavioral changes, merely a leaner execution path.

---
*PR created automatically by Jules for task [1698462274254453992](https://jules.google.com/task/1698462274254453992) started by @rfv-code*


rfv-code:
Closing as duplicate of #111.

#111 is the canonical for this optimization: 4 commits, and it carries the boundary-parity tests in `tests/connectors/db/policy_security.test.ts` (including `0889e2a`, which pins parity at the *position* level rather than the count, and `5dc273d`, which restored cases an earlier regen of this same PR dropped). This PR is one commit with no tests.

Measured per declaration in `src/connectors/db/lib/policy.ts` against the merge base:
- `_boundarySemicolons`: **changed identically** to #111 (`Set<number>` → `number[]`, `.add` → `.push`).
- `_splitStatements`: same change — drop `Array.from(...).sort()`, compare `b1[i] !== b2[i]` directly. Differs from #111 only in comment text.
- `Policy`, `_DELETE_KEYWORDS`, `_PRIVILEGE_KEYWORDS`: token-diffed to **punctuation only** — two redundant parens, three trailing commas. Prettier artifacts from the whole-file reformat that accounts for this PR reading as +108/-47 against #111's +9/-9.

**0 declarations exclusive to this PR.** If you think a case here is missing from #111, please cherry-pick it onto that branch.

PR #249, opened 2026-09-20
⚡ Bolt: Refactor _boundarySemicolons to array to remove O(N log N) overhead

rfv-code:
💡 **What:** Changed `_boundarySemicolons` to return `number[]` populated via `.push()` instead of using a `Set<number>`. Removed the Set-to-Array conversion and the subsequent `O(N log N)` `Array.sort()` calls in `_splitStatements`.
🎯 **Why:** While parsing SQL, statement boundary indices naturally increase monotonically (from 0 to `sql.length`). Because they are naturally distinct and sorted, storing them in a `Set` and then sorting them later incurs unnecessary hashing and sorting `O(N log N)` overhead.
📊 **Impact:** Reduces time complexity of splitting compound statements from `O(N log N)` to `O(N)`, completely eliminating `Set` insertion overhead and redundant array sorting.
🔬 **Measurement:** Verify tests run successfully using `pnpm test tests/connectors/db/policy.test.ts`. Inspect `src/connectors/db/lib/policy.ts` to confirm `_boundarySemicolons` now utilizes a standard array and `_splitStatements` evaluates the arrays directly without `.sort()`.

---
*PR created automatically by Jules for task [16434291824584203988](https://jules.google.com/task/16434291824584203988) started by @rfv-code*


rfv-code:
Closing as duplicate of #111.

After stripping comments and whitespace, `src/connectors/db/lib/policy.ts` on this branch hashes to `a664cd9ccc28a7b9899febb5925167aa75517d30` (24597 chars) — byte-identical to #111's copy. The only difference between the two branches is the wording of the explanatory comment above the boundary comparison loop.

#111 is the canonical: four commits against this branch's one, and it carries `tests/connectors/db/policy_security.test.ts`, which this branch does not touch. Both make the same change — `_boundarySemicolons` returns `number[]` instead of `Set<number>`, and `_splitStatements` drops the two `Array.from().sort()` calls because the indices are already monotonic.

PR #250, opened 2026-09-20
🛡️ Sentinel: [HIGH] Fix fetch timeout to cover body stream (Slowloris DoS)

rfv-code:
🚨 **Severity:** HIGH

💡 **Vulnerability:**
The `fetchWithCaps` helper in `src/toolkit/fetch_helper.ts` configured an `AbortController` timeout for the HTTP fetch request. However, `clearTimeout(timer)` was invoked in a `finally` block immediately after `await fetch(...)` resolved, meaning the timeout only protected the connection and header-read phase. The subsequent `reader.read()` loop reading the response body was completely unbounded by the timeout, exposing the system to Slowloris-style Denial-of-Service attacks where a malicious server trickles the response body indefinitely to exhaust resources.

🎯 **Impact:**
An attacker could cause the connector hub to hang indefinitely by responding slowly during the body-streaming phase of any `fetchWithCaps` call, potentially exhausting available connection pools or memory.

🔧 **Fix:**
Expanded the `try` block to encompass the entire body streaming loop (`reader.read()`) and array buffering. The `clearTimeout(timer)` call was shifted to the absolute end in the combined `finally` block, ensuring the timeout enforces the entire HTTP request lifecycle from connection to the final body byte.

✅ **Verification:**
- Ran the `fetch_helper.test.ts` suite.
- Re-ran the whole `vitest` suite.
- Formatted via Prettier.
- Kept the diff well under 50 lines.

---
*PR created automatically by Jules for task [4914723040401230958](https://jules.google.com/task/4914723040401230958) started by @rfv-code*


rfv-code:
Closing as duplicate of #112.

Verified against #112's tip, not its first commit. This branch is `ahead=1 behind=0` with one commit and zero test files; #112 carries seven commits plus `tests/toolkit/fetch_helper.test.ts`.

Token-diffing `src/toolkit/fetch_helper.ts` against the merge base and against #112, the only tokens exclusive to this branch are Prettier punctuation: a line split in `mergeSignals`, a split on the `mergeSignals(timeoutCtl.signal, ...)` call, and two trailing commas. The Slowloris fix itself — move the body read inside the `try` so `clearTimeout` only fires in the `finally` — is identical to #112's `f6718d3`.

Two things #112 has that this branch does not:
1. `8860521` aborts with `new DOMException("fetch_helper timeout", "AbortError")`. This branch keeps `new Error("fetch_helper timeout")`, so every caller's conventional `err.name === "AbortError"` check misclassifies a timeout as a generic network failure — and now that the timer stays armed through the body stream, that misclassification reaches a second code path.
2. The `fetchImpl` injection parameter and the `fetch_attachment.ts` delegation to `fetchWithCaps`. Without the delegation, `fetchAttachment` still buffers the whole body before the size check.

PR #251, opened 2026-09-21
🛡️ Sentinel: [CRITICAL] Fix DoS vulnerability in fetch_helper

rfv-code:
🚨 **Severity:** CRITICAL
💡 **Vulnerability:** The `fetchWithCaps` utility previously cleared its timeout immediately after the HTTP headers resolved (the initial `fetch` promise). The subsequent streaming body read (`reader.read()` loop) was unprotected by the timeout. A malicious or degraded server could keep the connection open and trickle bytes very slowly (a Slowloris attack), indefinitely blocking the event loop or consuming resources without triggering the timeout.
🎯 **Impact:** Denial of Service (DoS). An attacker could exhaust connector resources by returning large headers and then trickling the body, causing the agent to hang indefinitely on a single request.
🔧 **Fix:** Expanded the `try...finally` block that handles `clearTimeout(timer)` to encompass the entire body streaming loop, not just the initial `fetch` call. This ensures the timer remains active until all data is read or the cap is reached.
✅ **Verification:** Added a Vitest test (`aborts when the stream trickles slower than the timeout (Slowloris)`) that mocks a `ReadableStream` yielding one byte, waiting past the timeout, then yielding the next. The test verifies that the timeout correctly fires and throws an exception during the stream trickle. All tests pass successfully.

---
*PR created automatically by Jules for task [10743103061281713219](https://jules.google.com/task/10743103061281713219) started by @rfv-code*


rfv-code:
Closing as a duplicate of #112.

The substantive change here — moving `clearTimeout(timer)` into a `finally` that wraps the body-read loop, so the timeout stays armed through the stream — already landed on #112 in `f6718d3` ("keep timeout armed during body stream to close Slowloris"). The rest of this diff is a Prettier reformat of `fetch_helper.ts`.

#112 is the broader PR and is kept because it also:
- aborts with `new DOMException(…, "AbortError")` rather than a plain `Error`, so the conventional `err.name === "AbortError"` check no longer misclassifies a timeout as a network failure (`c04ec6f`);
- applies the same cap/timeout fix to `src/toolkit/fetch_attachment.ts`, which this branch leaves untouched;
- carries 4 new tests to this branch's 1. Its `keeps the timeout armed during the body stream (Slowloris)` test covers the same scenario as `aborts when the stream trickles slower than the timeout`, with a fully stalled body rather than a 50 ms drip.

One thing to flag rather than cherry-pick: this branch adds a **new 4724-line `pnpm-lock.yaml`**. This repo uses `package-lock.json`; that file should not be committed.

